PR #1, opened 2018-01-10
* Updated operations over FDataGrid objects to allow array and scalar operands.

vnmabus:


vnmabus:
Please, let me know if you find something undesirable about these changes.

fda/FDataGrid.py
```diff
@@ -262,57 +264,64 @@ def derivative(self, order=1):
         return FDataGrid(data_matrix, sample_points, self.sample_range,
                          names)
 
-    def __add__(self, other):
-        if not isinstance(other, FDataGrid):
-            return NotImplemented
+    def __check_same_dimensions(self, other):
         if self.data_matrix.shape[1] != other.data_matrix.shape[1]:
```
vnmabus: By the way, why only the second dimension is checked?

PR #5, opened 2018-09-30
Feature - shift registration

pablomm:
Shift registration features moved to the main repository.

Methods and changes added to **fda/basis.py**:
* Methods shift_registration, landmark_shift, shift and evaluate_shifted in FDataBasis
* Auxiliar function _check_extrapolation
* Method rescale in Basis and in the subclasses Fourier, BSplines and Monomial
* Added variable in Basis to define the default extrapolation

New module **fda/registration.py** with the methods to perform registration of FDataBasis objects.
Could be extend to FDataGrid if implements shift and evaluate_shifted.

Added script **shift_registration.py** with an example of an ideal case of shift registration.


fda/basis.py
```diff
@@ -1268,6 +1376,267 @@ def evaluate(self, eval_points, derivative=0):
...
+        else:
+            domain = self.domain_range
+
+        # Matrix of shifted values
+        _data_matrix = self.evaluate_shifted(tfine, shifts, ext=extrapolation)
+        _basis = self.basis.rescale(domain)
+
+        return FDataBasis.from_data(_data_matrix, tfine, _basis, **kwargs)
+
+    def shift_registration(self, maxiter=5, tol=1e-2, ext="default",
```
vnmabus: I am not sure it is a good idea to repeat functions as methods. I think that is more useful to have a few core small methods, and implement the functions using them, but we can discuss it. See https://github.com/numpy/numpy/issues/7452#issue-142974954

fda/registration.py
```diff
@@ -0,0 +1,274 @@
...
+
+    if len(landmarks) != fd.nsamples:
+        raise ValueError("landmark list ({}) must have the same length "
+                         "than the number of samples ({})"
+                         .format(len(landmarks), fd.nsamples))
+
+    landmarks = numpy.asarray(landmarks)
+
+    # Parses location
+    if location == 'minimize':
```
vnmabus: Another enumerated type would work better here.

fda/registration.py
```diff
@@ -0,0 +1,274 @@
...
+
+    else:
+        raise ValueError("Incorrect value of ext, should be 'default', 'basis'"
+                         ", 'periodic' or 'slice'.")
+
+    return periodic, extrapolation
+
+
+def shift_registration(fd, maxiter=5, tol=1e-2, ext="default", alpha=1,
+                       initial=[], tfine=[], shifts_array=False, **kwargs):
```
vnmabus: It is usually not a good idea in Python to have mutable objects as default values, as they are created only once, at function definition. I think in this case you do not modify the lists, but it is preferable to change them to tuples (immutable) or to None and create the lists in the body.

fda/basis.py
```diff
@@ -253,6 +284,24 @@ def penalty(self, derivative_degree=None, coefficients=None):
...
+            Args:
+                domain_range (tuple, optional): Definition of the interval where
+                    the basis defines a space. Defaults uses the same as the
+                    original basis.
+        """
+
+        if domain_range == None:
+            domain_range = self.domain_range
+        elif domain_range[0] >= domain_range[1]:
+            raise ValueError("The interval {} is not well-defined.".format(
```
vnmabus: As we are targeting only Python version 3.6 or above, I think we should use [f-strings](https://www.python.org/dev/peps/pep-0498/) as they are more legible than format.

fda/registration.py
```diff
@@ -4,38 +4,60 @@
...
-    elif ext == "periodic" or ext == 2:
-        extrapolation = 2
-        periodic = True
+        * default: uses the default method defined in the object to extrapolate.
+        * extrapolation: uses the extrapolated values directly.
+        * periodic: extends the domain range periodically.
+        * const: uses the boundary value.
+        * slice: avoids extrapolation restricting the domain.
+    """
+    default = 0
```
vnmabus: The values can be the strings. Do we want the numerical values for **anything**?

fda/registration.py
```diff
@@ -4,38 +4,60 @@
...
-        raise ValueError("Incorrect value of ext, should be 'default', 'basis'"
-                         ", 'periodic' or 'slice'.")
+        if isinstance(arg, ExtrapolationType): arg = arg.value
+        elif isinstance(arg, str): arg = arg.lower()
+
+        for ext in ExtrapolationType:
+
+            if arg == ext.value or arg == ext.name:
+
+                if ext.value == ExtrapolationType.default.value:
```
vnmabus: I think it is easier to write:
`if ext is ExtrapolationType.default:`

fda/registration.py
```diff
@@ -144,8 +166,8 @@ def shift_registration(fd, maxiter=5, tol=1e-2, ext="default", alpha=1,
     max_diff = tol + 1
     iter = 0
 
-    # Saves the array in all the interval if it will be adjusted to slice
-    if not periodic:
+    # Auxiliar array if the domain will be restricted
+    if extrapolation == ExtrapolationType.slice:
```
vnmabus: Comparison with singletons should be done using the `is` identity operator.

fda/basis.py
```diff
@@ -1514,7 +1484,7 @@ def shift(self, shifts, ext="default", tfine=[], **kwargs):
             domain = self.domain_range
 
         # Matrix of shifted values
-        _data_matrix = self.evaluate_shifted(tfine, shifts, ext=extrapolation)
+        _data_matrix = self.evaluate_shifted(tfine, shifts, ext=extrapolation.value)
```
vnmabus: Why not `extrapolation` but its value?

fda/registration.py
```diff
@@ -0,0 +1,274 @@
...
+This module contains the methods to perform the registration of
+functional data and related routines, in basis form as well in discretized form.
+
+"""
+
+import numpy
+import scipy.integrate
+
+
+def _check_extrapolation(fd, ext):
```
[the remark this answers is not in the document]
pablomm: This function is not needed after the changes added in  b5a30f6.

fda/registration.py
```diff
@@ -0,0 +1,274 @@
...
+    # Check how to extrapolate
+    # By default uses the value of the fd object
+    if ext == "default" or ext == 0:
+        ext = fd.default_extrapolation
+
+    if ext == "extrapolation" or ext == 1:
```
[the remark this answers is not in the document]
pablomm: Fixed in b5a30f6. Enumeration ExtrapolationType added in fda/registration.py.

fda/registration.py
pablomm: As I told you, I'm thinking of changing this argument and accepting a numeric value, a function to be called with landmark's vector or, by default, take 0.5(min(landmarks) + max(landmarks)) that minimizes the max shift.

fda/basis.py
pablomm: It's a little typo, I will fix it soon. I would like talking about if it's necessary evaluate_shifted to be a public function. If it would be a private function, the error control of the arguments could be omitted in order to improve performance.

fda/registration.py
pablomm: Fixed in 9c17e4d.

fda/registration.py
pablomm: Fixed in 9c17e4d.

fda/basis.py
pablomm: Changed in e492eda.

fda/registration.py
pablomm: Changed in a7c233a.

fda/registration.py
pablomm: Changed in a94db9f.

fda/basis.py
pablomm: Fixed in 9c17e4d.

fda/basis.py
pablomm: At the moment the repository is compatible with python 3.5, adding these changes would make it incompatible and travis would not compile correctly. 
https://github.com/GAA-UAM/fda/blob/c09ec2253f10d60157b4ceb453d3e19d53eed4a1/.travis.yml#L2-L3
I think it is not a good idea to add these changes until we decide which version to use.

fda/registration.py
pablomm: Changed in ffb48cd.

pablomm:
Examples of shift_registration and landmark_shift can be found in [this link](https://gist.github.com/pablomm/0009cba28d4c0eda925d9f4fd4137fdc).

pablomm:
Examples of interpolation of FDataGrid's can be found now on [gist](https://gist.github.com/pablomm/0009cba28d4c0eda925d9f4fd4137fdc).

PR #16, opened 2018-11-16
Feature/dataset improvements

vnmabus:
* Add the Medflies dataset
* Retrieve datasets by version instead of URL
* Import skdatasets only when necessary


fda/datasets/_real_datasets.py
```diff
@@ -1,6 +1,5 @@
 import numpy as np
 import rdata
```
pablomm: The `rdata` package is only used in `fetch_cran_dataset`, is there any reason to let the import in the module instead to moved it into the function, as `skdatasets`?

vnmabus: It is recommended to have the imports at the beginning unless there are powerful reasons for not doing that.

PR #23, opened 2018-11-22
Feature/fda graphics

amandaher:
Expansion of plot and scatter functions of FDataGrid to multiple dimensions on the image and up to two on the domain.
Fraiman & Muniz, Band and Modified Band Depths introduced.


fda/depth_measures.py
```diff
@@ -0,0 +1,173 @@
...
+    univariate_depth = np.zeros(fdgrid.shape)
+
+    if fdgrid.ndim_domain == 1:
+        for i in range(fdgrid.ndim_image):
+            for j in range(fdgrid.ncol):
+                column = fdgrid.data_matrix[:, j, i]
+                univariate_depth[:, j, i] = 1 - np.abs(0.5 - cumulative_distribution(column))
+    else:
+        for i in range(fdgrid.ndim_image):
+            for j in range(len(fdgrid.sample_points[1])):
```
vnmabus: As in the previous functions, maybe using `itertools` is more convenient, and allows to generalize to more dimensions. Maybe even you can vectorize the operation, making loops unnecessary.

fda/depth_measures.py
```diff
@@ -0,0 +1,173 @@
+"""Depth Measures Module.
+
+This module includes different methods to order functional data,
+from the center (larger values) outwards(smaller ones)."""
+
+import math
+import numpy as np
+from scipy.stats import rankdata
+from fda.grid import FDataGrid
```
vnmabus: Please, use `.grid` instead of `fda.grid` if you can. It will be easier to change the package name that way.

fda/grid.py
```diff
@@ -543,91 +547,151 @@ def concatenate(self, other):
                          self.dataset_label,
                          self.axes_labels)
 
-    def _set_labels(self, ax):
+    def _set_labels(self, fig):
```
vnmabus: This function previously could accept either `matplotib.pyplot` or an axes object, IIRC. Can it accept those now?

fda/grid.py
```diff
@@ -543,91 +547,151 @@ def concatenate(self, other):
...
+        """Arrange the layout of the figure.
+
+        Args:
+            fig (figure object): figure object to be arranged.
+
+        """
+        fig.tight_layout()
+        if self.dataset_label is not None:
+            st = fig.texts[0]
+            st.set_y(0.95)
```
vnmabus: I am not sure about the magic numbers. Why are they necessary?

fda/grid.py
```diff
@@ -543,91 +547,151 @@ def concatenate(self, other):
...
         """
         if self.dataset_label is not None:
-            try:
-                ax.set_title(self.dataset_label)
-            except AttributeError:
-                try:
-                    ax.title(self.dataset_label)
-                except AttributeError:
-                    pass
+            fig.suptitle(self.dataset_label, fontsize="x-large")
```
vnmabus: Maybe it would be useful to allow changing plot parameters, such as the fontsize.

fda/grid.py
amandaher: Previously it could only accept an axes object, now it can only accept a figure object since the purpose of this change is to expand it to more dimensions. There are as many axes objects as dimensions of the image, so if an axes object was accepted, the labels should be passed as arguments to avoid confusion.

fda/grid.py
amandaher: They are necessary because the figure is composed of several axes objects without taking into account if we want a common title for all of them (dataset_label). If this was not done, the title would  overlap with the axes on top.

fda/depth_measures.py
amandaher: In this case, the operation cannot be vectorized as np.unique (which is then used in cumulative_distribution function) only accepd 1D arrays.

fda/graphics.py
```diff
@@ -0,0 +1,212 @@
...
+import matplotlib.pyplot as plt
+import math
+from .depth_measures import *
+from .grid import FDataGrid
+
+__author__ = "Amanda Hernando Bernabé"
+__email__ = "amanda.hernando@estudiante.uam.es"
+
+
+def fdboxplot(fdgrid, fig=None, method=modified_band_depth, prob=[0.5], fullout=False, factor=1.5, barcol="blue",
```
vnmabus: I thought that we said that we would return a class containing all the information in the boxplot, and which can be plotted, so that retrieving the information is independent on the plot. Why is that not the case, at the end?

fda/clustering.py
```diff
@@ -0,0 +1,67 @@
...
+import scipy
+import math
+
+
+from .grid import FDataGrid
+
+__author__ = "Amanda Hernando Bernabé"
+__email__ = "amanda.hernando@estudiante.uam.es"
+
+from fda.depth_measures import *
```
manso92: Las dependencias están importadas dos veces

PR #26, opened 2018-12-02
Feature/fdata

pablomm:
Created a common class `FData` to treat `FDataBasis` and `FDataGrid` indistinctly for external methods.

The most important feature is that they both have a method for evaluating with the same api, which is
compatible with multidimensional objects, but to do this I had to change the `domain_range`  of `FDataBasis`, now is a list of axis like `FDataGrid`.

There are still things to do, for example, extrapolation, some basis operators, update documentation to new changes, or add doctests and more exhaustive unittests.

PR #29, opened 2018-12-16
Feature/inner product

manso92:


fda/basis.py
```diff
@@ -276,6 +276,12 @@ def copy(self):
         """Basis copy"""
         return copy.deepcopy(self)
 
+    def tofdatabasis(self):
```
vnmabus: It should be `to_basis` to be consistent with the FData API. I know that the name is confusing, and maybe it should be changed later, but it should be the same.

manso92: Ok

manso92:
Updated all the changes to pass the pull

fda/basis.py
```diff
@@ -1778,6 +1784,21 @@ def copy(self, *, basis=None, coefficients=None, dataset_label=None,
...
+            raise ValueError("Both Objects should have the same domain_range")
+        if isinstance(other, Basis):
+            other = other.to_basis()
+        inprodmat = [
+            scipy.integrate.quad(lambda x: self[i].evaluate([x]) * other[j].evaluate([x]), self.domain_range[0],
+                                 self.domain_range[1])[0] for j in range(0, other.nsamples) for i in
+                                 range(0, self.nsamples)]
+        return numpy.array(inprodmat).reshape((self.nsamples, other.nsamples))
+
+    def __getitem__(self, given):
```
pablomm: The function `__getitem__` was added in the pull request   #26 `feature/FData`.

fda/basis.py
```diff
@@ -1778,6 +1784,21 @@ def copy(self, *, basis=None, coefficients=None, dataset_label=None,
                           axes_labels=axes_labels, extrapolation=extrapolation,
                           keepdims=keepdims)
 
+    def inprod(self, other):
+        if self.domain_range != other.domain_range:
```
pablomm: After the creation of `FData` the `domain_range` is stored in an array with the domain ranges of each dimension, to treat `FDataBasis` and  `FDataGrid` uniformly.

fda/basis.py
manso92: I discard it to use that one

fda/basis.py
manso92: I've change it to only see to the first item as is used in

https://github.com/GAA-UAM/fda/blob/92c984cb42a505661716c26f4cd56a2f1e69e26f/fda/basis.py#L218

It will be updated in next steps of the development

manso92:
Some changes to update the request changes

fda/basis.py
pablomm: In the new version the attributes are not being preserved and they'll have the default values in the slice.

```
>>> fdbasis.extrapolation = "zeros"
>>> fdbasis.keepdims = True
>>> fdbasis
FDataBasis(
    basis=Monomial(domain_range=[array([0, 1])], nbasis=1),
    coefficients=[[0.5]
     [1. ]],
    dataset_label=None,
    axes_labels=None,
    extrapolation=<fda.extrapolation.FillExtrapolation object at 0x1a19846b00>,
    keepdims=True)
>>> fdbasis[0]
FDataBasis(
    basis=Monomial(domain_range=[array([0, 1])], nbasis=1),
    coefficients=[[0.5]],
    dataset_label=None,
    axes_labels=None,
    extrapolation=None,
    keepdims=False)

```
In the discarded version it was used the method copy to avoid this problem.
```
    return self.copy(coefficients=self.coefficients[given])
```
The case in which the key is an integer was treated separately in the discarded version, as in `FDataGrid`, due to this behaviour of the slice of numpy
```
>>> matrix.shape
(2, 3, 2)
>>> matrix[0].shape
(3, 2)
>>> matrix[0:1].shape
(1, 3, 2)
```
It is true that it is not necessary to do it currently and could be deleted, but if more dimensions are added it will be necessary to do it.

vnmabus: Then only the last one is being used. Only one of them should be kept.

PR #37, opened 2019-01-21
Fixed the creation of a BSpline function with nbasis, knots and order…

manso92:
Fixed a casuistic for Bspline creations

### DIFF
--- a/fda/basis.py
+++ b/fda/basis.py
@@ -739,7 +739,7 @@ class BSpline(Basis):
             knots = list(knots)
             knots.sort()
             if domain_range is None:
-                domain_range = (0, 1)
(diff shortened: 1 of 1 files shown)


fda/basis.py
```diff
@@ -738,6 +738,8 @@ def __init__(self, domain_range=None, nbasis=None, order=4, knots=None):
         else:
             knots = list(knots)
             knots.sort()
+            if domain_range is None:
+                domain_range = (0, 1)
```
vnmabus: It  should be:
```suggestion
                domain_range = (knots[0], knots[-1])
```
or I am missing something?

PR #42, opened 2019-01-28
Feature/plotting grid

amandaher:
FDataGrid ploting extended to more than one dimension on the image and up to two dimension on the domain.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -51,7 +51,7 @@ help:
 clean:
 	rm -rf $(BUILDDIR)/*
 	rm -rf auto_examples
-	rm -rf autosummary
+    rm -rf modules/autosummary
 
 .PHONY: html
 html:
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -51,7 +51,7 @@ help:
 clean:
 	rm -rf $(BUILDDIR)/*
 	rm -rf auto_examples
-    rm -rf modules/autosummary
+	rm -rf modules/autosummary
 
 .PHONY: html
 html:
--- a/examples/plot_extrapolation.py
+++ b/examples/plot_extrapolation.py
@@ -52,7 +52,7 @@ fd_bspline.dataset_label = "BSpline Basis"
 
 # Plot of diferent representations
 fig, ax = plt.subplots(2,2)
-fdgrid.plot(ax = [ax[0][0]])
+fdgrid.plot(ax=[ax[0][0]])
 fd_fourier.plot(ax[0][1])
 fd_monomial.plot(ax[1][0])
 fd_bspline.plot(ax[1][1])
@@ -96,7 +96,7 @@ ax[0][1].set_xticks([])
 
 
 # Plot objects in the domain range
-fdgrid.plot(ax = [ax[0][0]])
+fdgrid.plot(ax=[ax[0][0]])
 fd_fourier.plot(ax[0][1])
 fd_monomial.plot(ax[1][0])
 fd_bspline.plot(ax[1][1])
--- a/examples/plot_extrapolation.py
+++ b/examples/plot_extrapolation.py
@@ -1,6 +1,7 @@
 """
 Extrapolation
 =============
+
 Shows the usage of the different types of extrapolation.
 """
 
--- a/examples/plot_extrapolation.py
+++ b/examples/plot_extrapolation.py
@@ -1,7 +1,6 @@
 """
 Extrapolation
 =============
-
 Shows the usage of the different types of extrapolation.
 """
 
@@ -52,7 +51,7 @@ fd_bspline.dataset_label = "BSpline Basis"
 
 # Plot of diferent representations
 fig, ax = plt.subplots(2,2)
-fdgrid.plot(ax[0][0])
+fdgrid.plot(ax = [ax[0][0]])
 fd_fourier.plot(ax[0][1])
 fd_monomial.plot(ax[1][0])
 fd_bspline.plot(ax[1][1])
@@ -96,7 +95,7 @@ ax[0][1].set_xticks([])
 
 
 # Plot objects in the domain range
-fdgrid.plot(ax[0][0])
+fdgrid.plot(ax = [ax[0][0]])
 fd_fourier.plot(ax[0][1])
 fd_monomial.plot(ax[1][0])
 fd_bspline.plot(ax[1][1])
@@ -114,8 +113,8 @@ fd_bspline.plot(ax[1][1])
 
 plt.figure()
 
-# Evaluation of the grid
-# Extrapolation supplied in the evaluation
+# Evaluation of the grid
+# Extrapolation supplied in the evaluation
 values = fdgrid(t, extrapolation="periodic")
 
 plt.plot(t, values.T, linestyle='--')
@@ -138,7 +137,7 @@ plt.figure()
 # Other way to call the extrapolation, changing the default value
 fdgrid.extrapolation = "bounds"
 
-# Evaluation of the grid
+# Evaluation of the grid
 values = fdgrid(t)
 plt.plot(t, values.T, linestyle='--')
 
@@ -159,7 +158,7 @@ plt.title("Boundary extrapolation")
 
 plt.figure()
 
-# Evaluation of the grid filling with zeros
+# Evaluation of the grid filling with zeros
 values = fdgrid(t, extrapolation="zeros")
 plt.plot(t, values.T, linestyle='--')
 
--- a/fda/grid.py
+++ b/fda/grid.py
@@ -685,8 +685,14 @@ class FDataGrid(FData):
                                                         other.data_matrix),
                                                        axis=0))
 
-    def set_figure_and_axes(self):
+    def set_figure_and_axes(self, nrows, ncols):
         """Set figure and its axes.
+
+        Args:
+            nrows(int, optiona): designates the number of rows of the figure to plot the different dimensions of the
+                image. Only specified if fig and ax are None. ncols must be also be customized in the same call.
+            ncols(int, optiona): designates the number of columns of the figure to plot the different dimensions of the
+                image. Only specified if fig and ax are None. nrows must be also be customized in the same call.
 
         Returns:
             fig (figure object): figure object initialiazed.
@@ -700,8 +706,10 @@ class FDataGrid(FData):
         else:
             projection = '3d'
 
-        ncols = math.ceil(math.sqrt(self.ndim_image))
-        nrows = math.ceil(self.ndim_image / ncols)
+        if ncols is None and nrows is None:
+            ncols = math.ceil(math.sqrt(self.ndim_image))
+            nrows = math.ceil(self.ndim_image / ncols)
+
         for i in range(self.ndim_image):
             fig.add_subplot(nrows, ncols, i + 1, projection=projection)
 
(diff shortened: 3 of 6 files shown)


examples/plot_extrapolation.py
```diff
@@ -52,7 +52,7 @@
 
 # Plot of diferent representations
 fig, ax = plt.subplots(2,2)
-fdgrid.plot(ax[0][0])
+fdgrid.plot(ax = [ax[0][0]])
```
vnmabus: As of PEP8, named parameters are used without spaces:
```suggestion
fdgrid.plot(ax=[ax[0][0]])
```

examples/plot_extrapolation.py
```diff
@@ -96,7 +96,7 @@
 
 
 # Plot objects in the domain range
-fdgrid.plot(ax[0][0])
+fdgrid.plot(ax = [ax[0][0]])
```
vnmabus:
```suggestion
fdgrid.plot(ax=[ax[0][0]])
```

docs/Makefile
```diff
@@ -51,7 +51,7 @@ help:
 clean:
 	rm -rf $(BUILDDIR)/*
 	rm -rf auto_examples
-	rm -rf autosummary
+	rm -rf modules/autosummary
```
vnmabus: Why does this change?

fda/grid.py
```diff
@@ -677,92 +685,157 @@ def concatenate(self, other):
                                                         other.data_matrix),
                                                        axis=0))
 
-    def _set_labels(self, ax):
+    def set_figure_and_axes(self):
```
vnmabus: Why is this public?

fda/grid.py
```diff
@@ -677,92 +685,157 @@ def concatenate(self, other):
...
+
+        """
+        fig = plt.gcf()
+
+        if self.ndim_domain == 1:
+            projection = None
+        else:
+            projection = '3d'
+
+        ncols = math.ceil(math.sqrt(self.ndim_image))
```
vnmabus: Why the sqrt? I think I am missing something.

fda/grid.py
```diff
@@ -677,92 +685,157 @@ def concatenate(self, other):
...
+            plt.subplots_adjust(wspace=0.4)
+
+        if nrows > 1 and self.axes_labels is not None:
+            plt.subplots_adjust(hspace=0.4)
+
+        ax = fig.get_axes()
+
+        return fig, ax
+
+    def set_labels(self, fig = None, ax = None):
```
vnmabus: Why is this public?

vnmabus:
```suggestion
    def set_labels(self, fig=None, ax=None):
```

fda/grid.py
```diff
@@ -677,92 +685,157 @@ def concatenate(self, other):
...
+                for i in range(self.ndim_image):
+                    ax[i].set_xlabel(self.axes_labels[0])
+                    ax[i].set_ylabel(self.axes_labels[1])
+                    ax[i].set_zlabel(self.axes_labels[i + 2])
+            else:
+                for i in range(self.ndim_image):
+                    ax[i].set_xlabel(self.axes_labels[0])
+                    ax[i].set_ylabel(self.axes_labels[i + 1])
+
+    def plot(self, fig=None, ax = None, **kwargs):
```
vnmabus:
```suggestion
    def plot(self, fig=None, ax=None, **kwargs):
```

fda/grid.py
```diff
@@ -677,92 +685,157 @@ def concatenate(self, other):
...
-        return _plot
-
-    def scatter(self, ax=None, **kwargs):
+        if self.ndim_domain > 2:
+            raise NotImplementedError("Plot only supported for functional data"
+                                      "modeled in at most 3 dimensions.")
+        if fig is not None and ax is not None:
+            raise ValueError("fig and axes parameters cannot be passed as arguments at the same time.")
+
+        if fig != None and len(fig.get_axes()) != self.ndim_image:
```
vnmabus:
```suggestion
        if fig is not None and len(fig.get_axes()) != self.ndim_image:
```

fda/grid.py
```diff
@@ -677,92 +685,157 @@ def concatenate(self, other):
...
+            raise ValueError("Number of axes of the figure must be equal to"
+                             "the dimension of the image.")
+
+        if ax is not None and len(ax)!= self.ndim_image:
+            raise ValueError("Number of axes must be equal to the dimension of the image.")
+
+        if fig == None and ax == None:
```
vnmabus:
```suggestion
        if fig is None and ax is None:
```

fda/grid.py
```diff
@@ -677,92 +685,157 @@ def concatenate(self, other):
...
+            fig, ax = self.set_figure_and_axes()
+
+        if self.ndim_domain == 1:
+            for i in range(self.ndim_image):
+                ax[i].plot(self.sample_points[0], self.data_matrix[:, :, i].T, **kwargs)
```
vnmabus: What happens here if you pass fig?

fda/grid.py
```diff
@@ -677,92 +685,157 @@ def concatenate(self, other):
...
-                               **kwargs)
-        self._set_labels(ax)
-        return _plot
+        if self.ndim_domain > 2:
+            raise NotImplementedError("Plot only supported for functional data"
+                                      "modeled in at most 3 dimensions.")
+        if fig is not None and ax is not None:
+            raise ValueError("fig and axes parameters cannot be passed as arguments at the same time.")
+
+        if fig != None and len(fig.get_axes()) != self.ndim_image:
```
vnmabus:
```suggestion
        if fig is not None and len(fig.get_axes()) != self.ndim_image:
```

fda/grid.py
```diff
@@ -677,92 +685,157 @@ def concatenate(self, other):
...
+            raise ValueError("Number of axes of the figure must be equal to"
+                             "the dimension of the image.")
+
+        if ax is not None and len(ax)!= self.ndim_image:
+            raise ValueError("Number of axes must be equal to the dimension of the image.")
+
+        if fig == None and ax == None:
```
vnmabus:
```suggestion
        if fig is None and ax is None:
```

fda/grid.py
```diff
@@ -677,92 +685,157 @@ def concatenate(self, other):
...
-        if ax is None:
-            ax = matplotlib.pyplot.gca()
-        _plot = None
-        for i in range(self.nsamples):
-            _plot = ax.scatter(self.sample_points[0],
-                               self.data_matrix[i],
-                               **kwargs)
-        self._set_labels(ax)
-        return _plot
+        if self.ndim_domain > 2:
```
vnmabus: Not very important, but maybe the checks can be extracted to a function, as they are used twice (in `plot` and `scatter`).

In fact, the two functions are nearly identical, so maybe you can use a private function and parameterize the parts that are not common.

requirements.txt
```diff
@@ -1,4 +1,5 @@
 matplotlib
 numpy
 scipy
-setuptools
\ No newline at end of file
+setuptools
+sklearn
```
vnmabus: Technically you do not use Scikit-learn in this PR, but let it be as we will need it later.

docs/Makefile
amandaher: Because the directory autosummary is generated inside modules directory.

fda/grid.py
amandaher: It is also used in boxplot and MS-plot.

fda/grid.py
amandaher: ´ncols = math.ceil(math.sqrt(self.ndim_image))
nrows = math.ceil(self.ndim_image / ncols)´
The figure has as many axes as dimensions on the image. Imagine there are 4 dimensions,
we want a two graphs on top of another two, not 4 in line.

fda/grid.py
vnmabus: Then maybe it should not be a method, but a function in the plotting module.

docs/Makefile
vnmabus: Ok, so this corrects a previous bug, in that case, because you did not change the generation.

fda/grid.py
vnmabus: I do not agree. I think that the disposition should be parametrizable, at least, even if we preserve the default of making square grids.

amandaher: changed

fda/grid.py
amandaher: changed

fda/grid.py
amandaher: unification of checks done

fda/grid.py
```diff
@@ -700,22 +706,24 @@ def set_figure_and_axes(self):
         else:
             projection = '3d'
 
-        ncols = math.ceil(math.sqrt(self.ndim_image))
-        nrows = math.ceil(self.ndim_image / ncols)
+        if ncols is None and nrows is None:
```
vnmabus: If only one of those is `None`, the other can be inferred.

fda/grid.py
```diff
@@ -677,92 +685,196 @@ def concatenate(self, other):
...
+                for i in range(self.ndim_image):
+                    ax[i].set_xlabel(self.axes_labels[0])
+                    ax[i].set_ylabel(self.axes_labels[1])
+                    ax[i].set_zlabel(self.axes_labels[i + 2])
+            else:
+                for i in range(self.ndim_image):
+                    ax[i].set_xlabel(self.axes_labels[0])
+                    ax[i].set_ylabel(self.axes_labels[i + 1])
+
+    def _genereic_plotting_checks(self, fig=None, ax=None, nrows=None, ncols=None):
```
vnmabus:
```suggestion
    def _generic_plotting_checks(self, fig=None, ax=None, nrows=None, ncols=None):
```

fda/grid.py
```diff
@@ -677,92 +685,196 @@ def concatenate(self, other):
...
-        _plot = ax.plot(self.sample_points[0],
-                        numpy.transpose(numpy.squeeze(self.data_matrix,
-                                                      axis=2)),
-                        **kwargs)
-        self._set_labels(ax)
-
-        return _plot
-
-    def scatter(self, ax=None, **kwargs):
+        fig, ax = self._genereic_plotting_checks(fig, ax, nrows, ncols)
```
vnmabus:
```suggestion
        fig, ax = self._generic_plotting_checks(fig, ax, nrows, ncols)
```

fda/grid.py
```diff
@@ -677,92 +685,196 @@ def concatenate(self, other):
...
-        if ax is None:
-            ax = matplotlib.pyplot.gca()
-        _plot = None
-        for i in range(self.nsamples):
-            _plot = ax.scatter(self.sample_points[0],
-                               self.data_matrix[i],
-                               **kwargs)
-        self._set_labels(ax)
-        return _plot
+        fig, ax = self._genereic_plotting_checks(fig, ax, nrows, ncols)
```
vnmabus:
```suggestion
        fig, ax = self._generic_plotting_checks(fig, ax, nrows, ncols)
```

fda/grid.py
```diff
@@ -677,92 +685,196 @@ def concatenate(self, other):
...
+            raise ValueError("fig and axes parameters cannot be passed as arguments at the same time.")
+
+        if fig is not None and len(fig.get_axes()) != self.ndim_image:
+            raise ValueError("Number of axes of the figure must be equal to"
+                             "the dimension of the image.")
+
+        if ax is not None and len(ax)!= self.ndim_image:
+            raise ValueError("Number of axes must be equal to the dimension of the image.")
+
+        if ax is not None and fig is not None and (nrows is not None or ncols is not None):
```
vnmabus: Maybe:

```suggestion
        if (ax is not None or fig is not None) and (nrows is not None or ncols is not None):
```

fda/grid.py
```diff
@@ -677,92 +685,196 @@ def concatenate(self, other):
...
+            raise ValueError("The number of columns and/or number of rows of the figure, "
+                             "in which each dimension of the image is plotted, can only "
+                             "be customized in case fig is None and ax is None.")
+
+        if (nrows is not None and ncols is None) or (nrows is None and ncols is not None):
```
vnmabus: Why? We can use for the other value the minimum that works:
`ncols = math.ceil(self.ndim_image / nrows)`

fda/grid.py
amandaher: changed

amandaher: changed

PR #44, opened 2019-01-28
Feature/boxplot

amandaher:
Boxplot functionality added.


fda/boxplot.py
```diff
@@ -0,0 +1,585 @@
...
+    Args:
+        fdatagrid (FDataGrid): Object to be visualized.
+        fig (figure object, optional): figure over with the graphs are plotted in case ax is not specified.
+            If None and ax is also None, the figure is initialized.
+        ax (list of axis objects, optional): axis over where the graphs are plotted. If None, see param fig.
+        method (:ref:`depth measure <depth-measures>`, optional): Method used to order the data.
+            Defaults to :func:`modified band depth <fda.depth_measures.modified_band_depth>`.
+        prob (list of float, optional): List with float numbers (in the range from 1 to 0) that indicate which central regions to
+            represent. Defaults to [0.5] which represents the 50% central region.
+        fullout (boolean): If true, the entire curve of the outlier samples is shown. Defaulsts to False.
```
vnmabus: It is the opposite, actually. Also, the name is badly chosen. My first impression was that this parameter required the full curve to be outside the outlying envelope, instead of only a point, to be considered an outlier.

vnmabus:
```suggestion
        fullout (boolean): If true, the entire curve of the outlier samples is shown. Defaults to False.
```

amandaher: A sample is considered an outlier if any of its points falls outside the outlying envelope.
And I know that fullout should be the other way, but I took this parameter from the method implemented in R which has the following description:
fullout: logical for plotting outlying curves.  If FALSE (the default) then only the part
outside the box is plotted. If TRUE, complete outling curves are plotted

examples/plot_fdboxplot.py
```diff
@@ -0,0 +1,124 @@
...
+colors_by_climate = colormap(np.asarray(dataset["target"]) / (n_climates- 1))
+climate_colors = colormap(np.arange(n_climates) / (n_climates- 1))
+
+#Building the legend
+patches = []
+for i in range(n_climates):
+    patches.append(mpatches.Patch(color=climate_colors[i], label=climates[i]))
+
+#Plotting the curves and the legend with the desired colors.
+matplotlib.rcParams['axes.prop_cycle'] = cycler(color=colors_by_climate)
```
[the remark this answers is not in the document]
amandaher: Since the plot method of FDataGrid only calls once to matplotlib.pyplot.plot for all the curves in the FDataGrid, and this last method does not accept an array specifying the color of each sample, the only way to customize the colors is in this way.

fda/boxplot.py
```diff
@@ -0,0 +1,585 @@
...
+from .grid import FDataGrid
+
+__author__ = "Amanda Hernando Bernabé"
+__email__ = "amanda.hernando@estudiante.uam.es"
+
+def _generic_boxplot_checks(fdatagrid, fig, ax, method, nrows, ncols, colormap, factor):
+
+    fig, ax = fdatagrid.generic_plotting_checks(fig, ax, nrows, ncols)
+
+    if ((method != fraiman_muniz_depth) & (method != band_depth) & (method != modified_band_depth)):
```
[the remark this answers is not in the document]
amandaher: deleted

examples/plot_fdboxplot.py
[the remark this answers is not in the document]
vnmabus: Then we should make the plot method more customizable.

fda/fdata_boxplot.py
```diff
@@ -0,0 +1,627 @@
...
+        colormap (matplotlib.colors.LinearSegmentedColormap): Colormap from
+            which the colors to represent the central regions are selected.
+        central_regions (array, (fdatagrid.ndim_image * ncentral_regions, 2,
+            nsample_points)): contains the central regions.
+        outliers (array, (fdatagrid.ndim_image, fdatagrid.nsamples)):
+            contains the outliers
+        barcol (string): Color of the envelopes and vertical lines.
+        outliercol (string): Color of the ouliers.
+        mediancol (string): Color of the median.
+        fullout (boolean): If true, the entire curve of the outlier samples
```
vnmabus:
```suggestion
        fullout (boolean): If True, the entire curve of the outlier samples
```

fda/fdata_boxplot.py
```diff
@@ -0,0 +1,627 @@
...
+    def outliers(self):
+        return self._outliers
+
+    @property
+    def barcol(self):
+        return self._barcol
+
+    @barcol.setter
+    def barcol(self, color):
+        self._barcol = color
```
vnmabus: If the getter and setter are not doing additional work, is preferable to remove them and expose a public attribute (in Python).

fda/fdata_boxplot.py
vnmabus: In the example, by default the curves are shown, and setting this to True hides them. Thus, either the code or the documentation is wrong. I would also change the name for a more understandable one.

fda/fdata_boxplot.py
```diff
@@ -0,0 +1,627 @@
...
+            # median sample
+            ax[m].plot(self.fdatagrid.sample_points[0], self.median[m],
+                       color=self.mediancol, zorder=5)
+
+        self.fdatagrid.set_labels(fig, ax)
+
+        return fig, ax
+
+    def _repr_svg_(self):
+        plt.ioff()
```
vnmabus: I do not think this line is necessary.

```suggestion
```

fda/fdata_boxplot.py
```diff
@@ -0,0 +1,627 @@
...
+        fig, _ = self.plot()
+        output = StringIO()
```
vnmabus: I think this should be BytesIO.

fda/fdata_boxplot.py
```diff
@@ -0,0 +1,627 @@
...
+        fig.savefig(output, format='svg')
+        plt.close(fig)
+        plt.show()
```
vnmabus: No!! Ipython will display it if necessary.

```suggestion
```

fda/fdata_boxplot.py
```diff
@@ -0,0 +1,627 @@
...
+
+            # median sample
+            ax[m].plot(self.fdatagrid.sample_points[0], self.median[m],
+                       color=self.mediancol, zorder=5)
+
+        self.fdatagrid.set_labels(fig, ax)
+
+        return fig, ax
+
+    def _repr_svg_(self):
```
vnmabus: You should test this method in the examples, by using `fdboxplot` as the last line, instead of plotting explicitly.

fda/fdata_boxplot.py
```diff
@@ -0,0 +1,627 @@
...
+    @property
+    def central_envelope(self):
+        pass
+
+    @property
+    def outlying_envelope(self):
+        pass
+
+    @property
+    def colormap(self):
```
vnmabus: If the colormap is the same for all Boxplot types, you can implement it in the abstract class instead.

fda/fdata_boxplot.py
```diff
@@ -0,0 +1,627 @@
...
+    @colormap.setter
+    def colormap(self, value):
+        pass
+
+    @abstractmethod
+    def plot(self, fig=None, ax=None, nrows=None, ncols=None):
+        pass
+
+    @abstractmethod
+    def _repr_svg_(self):
```
vnmabus: This can be implemented here.

examples/plot_fdboxplot.py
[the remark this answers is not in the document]
amandaher: If a color parameter is added in the plot method, then the *kwargs should be checked in order not to contain another color parameter because it would crash if not. Or, it should be specified in the documentation and not check it?

fda/fdata_boxplot.py
amandaher: Do you mean to initialize the attribute in the _init_ function as self.barcol = "blue" and remove both 
setter and getter?

fda/fdata_boxplot.py
amandaher: If this line is not included when the object is interpreted, two images appear: the figure of the plot method and its default representation via _repr_svg_. This line prevents the first one from appearing.

examples/plot_fdboxplot.py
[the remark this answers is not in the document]
vnmabus: I am not sure of what you are trying to say.

fda/fdata_boxplot.py
vnmabus: Yes.

fda/fdata_boxplot.py
vnmabus: No, because you do not plot in this method.

amandaher:
``` def _repr_svg_(self):
        plt.ioff()
        fig, _ = self.plot()
        output = StringIO()
        fig.savefig(output, format='svg')
        plt.close(fig)
        return output.getvalue()``

File: fda/fdata_boxplot.py (hunk 7 above)
Comment (amandaher) [reply fda/fdata_boxplot.py]: I tried it with BytesIO and it returns the following:
```
FormatterWarning: image/svg+xml formatter returned invalid type <class 'bytes'> (expected <class 'str'>)
```
And no image is shown.

Comment (vnmabus) [reply fda/fdata_boxplot.py]: Ok

File: fda/fdata_boxplot.py (hunk 4 above)
Comment (amandaher) [reply fda/fdata_boxplot.py]: I have changed it in order the documentation and implementation make sense. Also I changed the name to outliers_repr but I do not like it much, any suggestions?

File: fda/fdata_boxplot.py (hunk 9 above)
Comment (amandaher) [reply fda/fdata_boxplot.py]: I am not very familiar with the Jupyter Notebooks generated by Sphinx but I noticed that only the outputs consisting in images of plotting methods are inserted. 
``` fdboxplot``` default representation is an image but it is not shown, maybe it is related to the fact of using StringIO and it is recognized as an string. I will investigate more.

File: fda/fdata_boxplot.py (hunk 4 above)
Comment (vnmabus) [reply fda/fdata_boxplot.py]: `show_full_outliers`?

File: examples/plot_fdboxplot.py (hunk 2 above)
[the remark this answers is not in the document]
Comment (amandaher) [reply examples/plot_fdboxplot.py]: The implementation of the plot method for FDataGrid:
```
    def plot(self, fig=None, ax=None, nrows=None, ncols=None, **kwargs):
    
        fig, ax = self.generic_plotting_checks(fig, ax, nrows, ncols)

        if self.ndim_domain == 1:
            for i in range(self.ndim_image):
                ax[i].plot(self.sample_points[0], self.data_matrix[:, :, i].T, **kwargs)
        else:
            X = self.sample_points[0]
            Y = self.sample_points[1]
            X, Y = numpy.meshgrid(X, Y)
            for i in range(self.ndim_image):
                for j in range(self.nsamples):
                    ax[i].plot_surface(X, Y, numpy.squeeze(self.data_matrix[j, :, :, i]).T, **kwargs)

        self.set_labels(fig, ax)

        return fig, ax
```
If we want to accept a color parameter that includes a color for all the curves, the resulting method would be:
```
    def plot(self, fig=None, ax=None, nrows=None, ncols=None, color = None, **kwargs):
    
        fig, ax = self.generic_plotting_checks(fig, ax, nrows, ncols)
      
        if color is not None:
            check_kwargs

        if self.ndim_domain == 1:
            for i in range(self.ndim_image):
                for j in range(self.nsamples):
                    ax[i].plot(self.sample_points[0], self.data_matrix[j, :, i].T, color = color[j], **kwargs)
        else:
            X = self.sample_points[0]
            Y = self.sample_points[1]
            X, Y = numpy.meshgrid(X, Y)
            for i in range(self.ndim_image):
                for j in range(self.nsamples):
                    ax[i].plot_surface(X, Y, numpy.squeeze(self.data_matrix[j, :, :, i]).T, **kwargs)

        self.set_labels(fig, ax)

        return fig, ax
```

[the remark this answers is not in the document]
Comment (amandaher) [reply examples/plot_fdboxplot.py]: The implementation of the plot method for FDataGrid:
```
    def plot(self, fig=None, ax=None, nrows=None, ncols=None, **kwargs):
    
        fig, ax = self.generic_plotting_checks(fig, ax, nrows, ncols)

        if self.ndim_domain == 1:
            for i in range(self.ndim_image):
                ax[i].plot(self.sample_points[0], self.data_matrix[:, :, i].T, **kwargs)
        else:
            X = self.sample_points[0]
            Y = self.sample_points[1]
            X, Y = numpy.meshgrid(X, Y)
            for i in range(self.ndim_image):
                for j in range(self.nsamples):
                    ax[i].plot_surface(X, Y, numpy.squeeze(self.data_matrix[j, :, :, i]).T, **kwargs)

        self.set_labels(fig, ax)

        return fig, ax
```
If we want to accept a color parameter that includes a color for all the curves, the resulting method would be:
```
    def plot(self, fig=None, ax=None, nrows=None, ncols=None, color = None, **kwargs):
    
        fig, ax = self.generic_plotting_checks(fig, ax, nrows, ncols)
      
        if color is not None:
            check_kwargs

        if self.ndim_domain == 1:
            for i in range(self.ndim_image):
                for j in range(self.nsamples):
                    ax[i].plot(self.sample_points[0], self.data_matrix[j, :, i].T, color = color[j], **kwargs)
        else:
            X = self.sample_points[0]
            Y = self.sample_points[1]
            X, Y = numpy.meshgrid(X, Y)
            for i in range(self.ndim_image):
                for j in range(self.nsamples):
                    ax[i].plot_surface(X, Y, numpy.squeeze(self.data_matrix[j, :, :, i]).T, **kwargs)

        self.set_labels(fig, ax)

        return fig, ax
```

File: fda/fdata_boxplot.py (hunk 7 above)
Comment (amandaher) [reply fda/fdata_boxplot.py]: I changed it to BytesIO with a coding transformation.

File: fda/fdata_boxplot.py (hunk 9 above)
Comment (amandaher) [reply fda/fdata_boxplot.py]: If you go to the `Examples which don’t produce image output` section in https://sphinx-gallery.readthedocs.io/en/latest/auto_examples/index.html, you can see that unless the `print` method is used no output is shown, and the print function calls the _str_ method, so this functionality cannot be shown in the examples :( I left it with the plot method to show the images and introduced a comment at the end.

Comment (vnmabus) [reply fda/fdata_boxplot.py]: Have you tried the `display` function provided by IPython?

Comment (amandaher) [reply fda/fdata_boxplot.py]: The ´display´ function does not work neither, I have come to https://github.com/sphinx-gallery/sphinx-gallery/issues/396 but I do not understand it very much, maybe you could clear up something. Also maybe it can be resolved using ´Custom Scrapers´ explained in https://sphinx-gallery.readthedocs.io/en/latest/configuration.html.

Comment (vnmabus) [reply fda/fdata_boxplot.py]: Ok, leave it like that for now.

File: fda/fdata_boxplot.py [hunk 12]
```diff
@@ -0,0 +1,572 @@
...
+    depending on the dimensions of the domain, 1 or 2 respectively.
+
+    It forces to both classes, Boxplot and SurfaceBoxplot to conain at least the median,
+    central and outlying envelopes and a colormap for their graphical representation,
+    obtained calling the plot method.
+
+    """
+    @abstractmethod
+    def __init__(self, factor=1.5):
+        if factor < 0:
```
vnmabus: Maybe factor should be stored, and exposed as a read-only property.

fda/fdata_boxplot.py
```diff
@@ -0,0 +1,572 @@
...
+        median (array, (fdatagrid.ndim_image, lx, ly)): contains
+            the median/s.
+        central_envelope (array, (fdatagrid.ndim_image, 2, lx, ly)):
+            contains the central envelope/s.
+        outlying_envelope (array,(fdatagrid.ndim_image, 2, lx, ly)):
+            contains the outlying envelope/s.
+        colormap (matplotlib.colors.LinearSegmentedColormap): Colormap from
+            which the colors to represent the central regions are selected.
+        boxcol (string): Color of the ouliers.
+        outcol (string): Color of the median.
```
vnmabus: Should it not be mediancol?

fda/fdata_boxplot.py
```diff
@@ -0,0 +1,572 @@
...
+    Example:
+        Function :math:`f : \mathbb{R}\longmapsto\mathbb{R}`.
+
+        >>> data_matrix = [[1, 1, 2, 3, 2.5, 2], [0.5, 0.5, 1, 2, 1.5, 1], [-1, -1, -0.5, 1, 1, 0.5],
+        ...                [-0.5, -0.5, -0.5, -1, -1, -1]]
+        >>> sample_points = [0, 2, 4, 6, 8, 10]
+        >>> fd = FDataGrid(data_matrix, sample_points, dataset_label="dataset", axes_labels=["x_label", "y_label"])
+        >>> Boxplot(fd) # doctest: +IGNORE_RESULT
+
+    """.replace('+IGNORE_RESULT', '+ELLIPSIS\n<...>')
```
vnmabus: ???

fda/fdata_boxplot.py
amandaher: done

fda/fdata_boxplot.py
amandaher: changed

PR #48, opened 2019-02-19
Fix/fourier evaluation

manso92:
Updated Fourier evaluation and its tests.


manso92:
Yes, the coefficients used are the result of the evaluation on r library

PR #49, opened 2019-02-21
New Lfd object for linear differential operator

manso92:
I've developed a new class for lineal differential operators.


fda/lfd.py
```diff
@@ -0,0 +1,91 @@
...
+                else:
+                    raise ValueError("FDataBasis objects in the list has not the same domain_range")
+            else:
+                raise ValueError("The elements of the list are neither integers or FDataBasis objects")
+
+        elif isinstance(param, FDataBasis):
+            self.nderiv = param.nsamples
+            self.bwtlist = param.to_list()
+
+        elif isinstance(param, int):
```
vnmabus: Now with the code in front of me I understand better why you chose two parameters instead of one, and I think that I agree with that approach. One would expect that f(i) and f([i]) are not very different.

Thus, it is better that the constructor accepts two parameters, as you said it did before: `coefficients`, and `order`. The second can be deduced from the first one if missing, and the first one would be `(0, 0, 0, ..., 0, 1)` if missing. The user MUST give only one of them.

Sorry for making you redo your work.

fda/lfd.py
```diff
@@ -0,0 +1,91 @@
...
+
+        if isinstance(param, list):
+            if all(isinstance(n, int) for n in param):
+                self.nderiv = len(param)
+                self.bwtlist = (FDataBasis(Constant(domain_range), np.array(param).reshape(-1, 1)).to_list()
+                                if self.nderiv != 0 else [])
+            elif all(isinstance(n, FDataBasis) for n in param):
+                if all([_same_domain(param[0].domain_range, x.domain_range) for x in param]):
+                    self.nderiv = len(param)
+                    self.bwtlist = param
```
vnmabus: Maybe we should check that they have only one sample.

fda/lfd.py
```diff
@@ -0,0 +1,91 @@
...
+                  If its a list, of integers or FDataBasis objects, it will be taken as the weight functions
+                  for the derivatives.
+                  If its a FDataBasis object, each sample will be a weight function and nderiv will be the number
+                  of samples
+
+            domain_range (tuple or list of tuples, optional): Definition of the
+                interval where the weight functions are defined. Defaults to (0,1).
+        """
+
+        if isinstance(param, list):
```
vnmabus: It is probably better if we iterate over, and do not assume a list. It should also do the right think given an FDataBasis.

fda/lfd.py
```diff
@@ -0,0 +1,91 @@
...
+            if param < 0:
+                raise ValueError("The deriv order must be non-negative.")
+
+            self.nderiv = param
+            self.bwtlist = [FDataBasis(Constant(domain_range), 0) for _ in range(param)] if self.nderiv != 0 else []
+
+        else:
+            raise ValueError("Argument is neither a non-negative integer or list")
+
+    def __repr__(self):
```
vnmabus: Maybe a `_repr_latex_` method can also be provided.

fda/lfd.py
```diff
@@ -0,0 +1,91 @@
...
+    def __eq__(self, other):
+        """Equality of Lfd objects"""
+        return (self.nderiv == other.nderic
+                and all(self.bwtlist[i] == other.bwtlist[i] for i in range(self.nderiv)))
+
+    @property
+    def domain_range(self):
+        """Definition range."""
+
+        return None if self.nderiv == 0 else self.bwtlist[0].domain_range
```
vnmabus: Why `None`? The user could have specified a value, and we also have a default one.

fda/lfd.py
```diff
@@ -0,0 +1,91 @@
...
+        Lx(t) = b_0(t) x(t) + b_1(t) x'(x) + \dots + b_{n-1}(t) d^{n-1}(x(t)) + b_n(t) d^n(x(t))
+
+    Attributes:
+        nderiv (int): the order of the operator. It's the n coefficient in the equation above.
+
+        bwtlist (list):  A FDataBasis objects list of length either nderiv
+
+    """
+
+    def __init__(self, param=0, domain_range=(0, 1)):
```
vnmabus: If `domain_range` is not given, but a FDataBasis is, we could deduce it. Also we should store it for the 0 order case.

fda/lfd.py
manso92: But for a latex repr we must have the latex repr of basis and FDataBasis objects. I think is better to open an issue and implement them.

If your agree I'll open it and I'll fix it

vnmabus: I agree.

fda/lfd.py
```diff
@@ -0,0 +1,106 @@
...
+    def __init__(self, order=None, weights=None, domain_range=(0, 1)):
+        """Lfd Constructor. You have to provide one of the two first
+           parameters. It both are provided, it will raise an error
+
+        Args:
+            order (int, optional): the order of the operator. It's the n
+                  coefficient in the equation above.
+
+            weights (list, optional): A FDataBasis objects list of length
+                     either nderiv
```
vnmabus: Incomplete sentence.

fda/lfd.py
```diff
@@ -0,0 +1,106 @@
...
+                self.order = len(weights) - 1
+                self.weights = (FDataBasis(Constant(domain_range),
+                                           np.array(weights)
+                                           .reshape(-1, 1)).to_list())
+
+            elif all(isinstance(n, FDataBasis) for n in weights):
+                if all([_same_domain(weights[0].domain_range,
+                                     x.domain_range) and x.nsamples == 1 for x
+                        in weights]):
+                    self.order = len(weights)
```
vnmabus: `len(weights)` - 1 maybe?

fda/lfd.py
```diff
@@ -0,0 +1,106 @@
...
+        bwtliststr = ""
+        for i in range(self.order):
+            bwtliststr = bwtliststr + "\n" + self.weights[i].__repr__() + ","
+
+        return (f"{self.__class__.__name__}("
+                f"\nnderiv={self.order},"
+                f"\nbwtlist=[{bwtliststr[:-1]}]"
+                f"\n)").replace('\n', '\n    ')
+
+    def __str__(self):
```
vnmabus: The default implementation already does that.

PR #54, opened 2019-02-26
Feature/fda srvf

pablomm:
Elastic registration methods implemented and a module with metrics was created.


deps/fdasrsf/DynamicProgrammingQ2.c
```diff
@@ -0,0 +1,31 @@
...
+#include "dp_grid.h"
+
+void DynamicProgrammingQ2(double *Q1, double *T1, double *Q2, double *T2, int m1, int n1, int n2,
+double *tv1, double *tv2, int n1v, int n2v, double *G, double *T, double *size, double lam1){
+  int *idxv1 = 0;
+  int *idxv2 = 0;
+  double *E = 0; /* E[ntv1*j+i] = cost of best path to (tv1[i],tv2[j]) */
+  int *P = 0; /* P[ntv1*j+i] = predecessor of (tv1[i],tv2[j]) along best path */
+
+  idxv1=(int*)malloc((n1v)*sizeof(int));
```
vnmabus: Those casts are unnecessary in C.

deps/fdasrsf/dp_grid.c
```diff
@@ -0,0 +1,284 @@
+#include <stdio.h>
```
vnmabus: I was thinking if we could standardize a C format style and apply it before the first commit. In particular, I would like 4 spaces (or a tab) per indentation level, spaces around operators, no spaces around parentheses...

deps/fdasrsf/dp_nbhd.h
```diff
@@ -0,0 +1,77 @@
+#ifndef DP_NBHD_H
+#define DP_NBHD_H 1
+
+#ifndef DP_NBHD_DIM
+#define DP_NBHD_DIM 7 
+#endif
+
+#if DP_NBHD_DIM == 17
+int dp_nbhd[][2] = {
```
vnmabus: OMG, what the hell is this??

deps/fdasrsf/dp_grid.c
pablomm: Done in 396600a.

deps/fdasrsf/DynamicProgrammingQ2.c
pablomm: Deleted in 396600a.

deps/fdasrsf/dp_nbhd.h
pablomm: I didn't change the original code following the next rule 😆 

image

In `dp_grid.c` a grid it is explored evaluating an energy function, this file contains the coordinates of the grid flattened. I guess they did it for performance reasons.

By the moment we only use one algorithm, but the original code contains more modules which uses this grid (maybe in a future they could be implemented), I guess this is the reason why the grid coordinates are in a global variable in a header.

vnmabus: But why the arbitrary limits? Why not providing an inline function to compute it instead? The performance should be almost the same (or even better, as you are playing nicely with the cache). This code looks COMPLETELY awful.

deps/fdasrsf/DynamicProgrammingQ2.c
```diff
@@ -0,0 +1,37 @@
...
+                          int m1, int n1, int n2,
+                          double* tv1, double* tv2, int n1v, int n2v, double* G,
+                          double* T, double* size, double lam1)
+{
+    int* idxv1;
+    int* idxv2;
+    double* E; /* E[ntv1*j+i] = cost of best path to (tv1[i],tv2[j]) */
+    int* P; /* P[ntv1*j+i] = predecessor of (tv1[i],tv2[j]) along best path */
+
+    idxv1 = malloc((n1v)*sizeof(int));
```
vnmabus: Please remove the unnecessary parentheses, and put spaces around the operators. I also prefer `sizeof(*idxv1)` instead of `sizeof(int)` to guarantee that the size is always ok.

deps/fdasrsf/DynamicProgrammingQ2.c
```diff
@@ -0,0 +1,37 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include "dp_grid.h"
+
+void DynamicProgrammingQ2(double* Q1, double* T1, double* Q2, double* T2,
+                          int m1, int n1, int n2,
+                          double* tv1, double* tv2, int n1v, int n2v, double* G,
+                          double* T, double* size, double lam1)
+{
+    int* idxv1;
```
vnmabus: I usually prefer to declare the variables when they are first assigned to, if possible. This can be done in C99 and C11 standards.

deps/fdasrsf/dp_grid.c
```diff
@@ -0,0 +1,257 @@
...
+    double* tv1, int* idxv1, int ntv1,
+    double* tv2, int* idxv2, int ntv2,
+    double* W, double lam)
+{
+    int sr, sc;  /* source row and column */
+    int tr, tc;  /* target row and column */
+    int l1, l2, l3;  /* for multidimensional array mapping */
+    int i;
+
+    for (i=0; i<ntv1*ntv2*ntv1*ntv2; W[i++]=1e6);
```
vnmabus: We should put magic numbers in `#define`'s. Also, spaces around operators.

deps/fdasrsf/dp_grid.c
```diff
@@ -0,0 +1,257 @@
...
+     *  (j,i,l,k) :--> j*ntv1*ntv2*ntv1 +
+     *                 i*ntv2*ntv1 +
+     *                 l*ntv1 +
+     *                 k
+     */
+    l1 = ntv1 * ntv2 * ntv1;
+    l2 = ntv2 * ntv1;
+    l3 = ntv1;
+
+    for (tr=1; tr<ntv2; ++tr) {
```
vnmabus: Spaces also around assign and comparison operators. Using `++` prefix notation indicates that this guy comes from a `C++` background, which may explain why the casts were used in `malloc` and `calloc`.

deps/fdasrsf/dp_grid.c
```diff
@@ -0,0 +1,257 @@
...
+        for (tc=1; tc<ntv1; ++tc) {
+            for (i=0; i<DP_NBHD_COUNT; ++i) {
+                sr = tr - dp_nbhd[i][0];
+                sc = tc - dp_nbhd[i][1];
+
+                if (sr < 0 || sc < 0) continue;
+
+                /* grid(sr,sc,tr,tc) */
+                W[sr*l1+sc*l2+tr*l3+tc] =
```
vnmabus: Spaces around operators.

deps/fdasrsf/dp_grid.c
```diff
@@ -0,0 +1,257 @@
...
+    double w, cand_cost;
+    int i;
+
+    E[0] = 0.0;
+    for (i=1; i<ntv1; E[i++]=1e6);
+    for (i=1; i<ntv2; E[ntv1*i++]=1e6);
+
+    for (tr=1; tr<ntv2; ++tr) {
+        for (tc=1; tc<ntv1; ++tc) {
+            E[ntv1*tr + tc] = 1e6;
```
vnmabus: That magic number again.

deps/fdasrsf/dp_grid.c
```diff
@@ -0,0 +1,257 @@
...
+    double* tv2, int* idxv2, int ntv2,
+    double* E, int* P, double lam)
+{
+    int sr, sc;  /* source row and column */
+    int tr, tc;  /* target row and column */
+    double w, cand_cost;
+    int i;
+
+    E[0] = 0.0;
+    for (i=1; i<ntv1; E[i++]=1e6);
```
vnmabus: Why so many `for` for the initialization of the same array?

deps/fdasrsf/dp_grid.c
```diff
@@ -0,0 +1,257 @@
...
+    rslope = sqrt(slope);
+
+    while (t1 < b && t2 < d) {
+        if (Q1idx > nsamps1-2 || Q2idx > nsamps2-2) break;
+
+        /* Find endpoint of current interval */
+        t1nextcand1 = T1[Q1idx+1];
+        t1nextcand2 = a + (T2[Q2idx+1]-c) / slope;
+
+        if (fabs(t1nextcand1-t1nextcand2) < 1e-6) {
```
vnmabus: Magic number!!

deps/fdasrsf/dp_nbhd.h
vnmabus: If we have proper tests, we should be able to change it safely.

deps/fdasrsf/optimum_reparamN.pyx
```diff
@@ -0,0 +1,271 @@
+# cython: language_level=2
```
vnmabus: Why are we using Python 2 here? Maybe this is why the output of numpy was the traditional one, as new versions have dropped Python 2 support.

docs/conf.py
```diff
@@ -55,6 +55,17 @@
...
 
+doctest_global_setup = '''
+try:
+    import fda
+    from fda import *
+    from fda.depth_measures import *
+    import numpy
+    numpy.set_printoptions(legacy='1.13')
+except:
+    pass
```
vnmabus: Is this a sensible thing to do?

deps/fdasrsf/dp_grid.c
pablomm: ``E`` is a 2-D array where it is computed the energy function. It should be initialised with E[0, 1:]= ∞ and E[1:, 0]=∞ and E[0,0]=0.

deps/fdasrsf/dp_grid.c
pablomm: Constant INFTY and spaces added in 36dbe63.

deps/fdasrsf/dp_grid.c
pablomm: Done in 36dbe63.

deps/fdasrsf/dp_grid.c
pablomm: Changed in 36dbe63.

deps/fdasrsf/dp_grid.c
pablomm: Added TOL in 36dbe63.

docs/conf.py
pablomm: try/except deleted in de19847.

deps/fdasrsf/optimum_reparamN.pyx
pablomm: Changed to python3 in de19847.

deps/fdasrsf/DynamicProgrammingQ2.c
pablomm: Done in de19847.

deps/fdasrsf/dp_nbhd.h
pablomm: Tests were added in de19847. 
Looking at the coordinate lists in detail I realised that they are not simply the Cartesian product, for optimisation reasons it is not necessary to evaluate all the points, which complicates the dynamic generation of the list with an inline function.

These are the coordinates in the list (colours indicate different sizes):
Figure_6

Maybe the list may be initialised at the beginning of the algorithm or we can use always the same size.

fda/registration/_registration_utils.py
```diff
@@ -187,3 +211,118 @@ def mse_decomposition(original_fdata, registered_fdata, warping_function=None,
...
+
+    data_matrix = numpy.empty((fdatagrid.nsamples, len(eval_points)))
+
+    for i in range(fdatagrid.nsamples):
+        data_matrix[i] = PchipInterpolator(y[i], eval_points)(eval_points)
+
+    return fdatagrid.copy(data_matrix=data_matrix, sample_points=eval_points)
+
+
+def _normalize_scale(t, a=0, b=1):
```
vnmabus: Should we use the domain_range, maybe? It is possible for the original data to already be in the desired interval, but without observations exactly at the endpoints.

fda/metrics.py
```diff
@@ -0,0 +1,502 @@
...
+        raise ValueError("Not implemented for image with "
+                         "dimension greater than 1")
+
+    # Computes the norm, approximating the integral with Simpson's rule.
+    return scipy.integrate.simps(numpy.abs(fdatagrid.data_matrix[..., 0]) ** p,
+                                 x=fdatagrid.sample_points
+                                 ) ** (1 / p)
+
+
+def metric(fdatagrid, fdatagrid2, norm=norm_lp, **kwargs):
```
vnmabus: Not all metrics come from a norm. I know that this function is from Miguel, but maybe we should change its name. In fact, I would split it into two functions:
- A small function that returns a distance given a norm, such as:
```python
def distance_from_norm(norm, **kwargs):
    def norm_distance(fdata1, fdata2):
        return norm(fdata1 - fdata2, **kwargs)
    
    norm_distance.__name__ = f"{norm.__name__}_distance"
    return norm_distance
```
- A function `pairwise_distances` that receives a distance (not a norm) and computes the pairwise distances.

Then each distance function should only need to care about distance between individual functions, and the pairwise distances  could be delegated to this one.

examples/plot_pairwise_alignment.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+###############################################################################
+# We can perform the pairwise of multiple curves at once. We can use a single
+# curve as template to align a set of samples to it or a set of
+# templates to make the alignemnt the two sets.
+#
+#
+# In the elastic registration example it is shown the alignment of multiple
+# curves to the same template is shown.
```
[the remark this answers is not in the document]
pablomm: a084ddf fix it.

deps/fdasrsf/dp_grid.c
```diff
@@ -0,0 +1,258 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <math.h>
+#include "dp_nbhd.h"
+#include "dp_grid.h"
+
+#define INFTY 1e6
```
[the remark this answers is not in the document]
pablomm: Changed to INFINITY in a084ddf.

deps/fdasrsf/optimum_reparamN.pyx
```diff
@@ -0,0 +1,290 @@
+# cython: language_level=3
+
+cimport cDPQ
+import numpy as np
+from numpy.linalg import norm
+
+cimport numpy as np
+from cpython cimport array
+
+def coptimum_reparamN(np.ndarray[double, ndim=1, mode="c"] mq,
```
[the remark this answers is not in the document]
pablomm: Module cleared in a084ddf.

pablomm:
> Now that we know with ones are missing, you can omit entirely the array. The loops can be redone skipping the redundant points. If you want to encapsulate the logic, you can make a function like:
`int next_grid_point(int x, int y, int * next_x, int * next_y, size_t grid_size)`
which returns 1 if there is not any next point and 0 otherwise.

_Originally posted by @vnmabus in https://github.com/GAA-UAM/fda/pull/54_

I think that this approach is too expensive, to align a sample the grid it is used `n^2`, where `n` is the number of sample points. In a084ddf I changed the algorithm to generate the grid at the beginning and it is passed as a local variable.
https://github.com/GAA-UAM/fda/blob/a084ddff984396805bcab68f6ebc46c67a703533/deps/fdasrsf/dp_grid.h#L147-L158
https://github.com/GAA-UAM/fda/blob/a084ddff984396805bcab68f6ebc46c67a703533/deps/fdasrsf/dp_grid.c#L339

fda/registration/_registration_utils.py
pablomm: Change in cb0dc3d. This is an internal function, but the API of the public function was changed, to use by default the domain range. 
https://github.com/GAA-UAM/fda/blob/ff243ad6ceae97896a5364f9b5136bb6d7e88b16/fda/registration/_registration_utils.py#L306

fda/metrics.py
pablomm: Changes added in 0fc7096. Now the distances will receive two FDatas with one sample and returns a float with the distance. Test for the metrics were added too.

Issue #53, opened 2019-02-25
First plot messes figure and subplots


vnmabus:
The first plot of a multivariate functional object is messed up. Subsequent plots are ok.

Issue #55, opened 2019-02-26
Bug in plot function


pablomm:
When a FData with an unidimensional domain object is plotted after other plot a matplotlib deprecated warning is obtained, but the result is correct.

```
>>> import matplotlib.pyplot as plt; import fda
>>> a = fda.datasets.make_multimodal_samples(ndim_domain=1)
>>> a.plot()
(<Figure size 640x480 with 1 Axes>, [<matplotlib.axes._subplots.AxesSubplot object at 0x1c1133a3c8>])
>>> a.plot()
/Users/pablomm/anaconda3/lib/python3.6/site-packages/matplotlib/cbook/deprecation.py:107: MatplotlibDeprecationWarning: Adding an axes using the same arguments as a previous axes currently reuses the earlier instance.  In a future version, a new instance will always be created and returned.  Meanwhile, this warning can be suppressed, and the future behavior ensured, by passing a unique label to each axes instance.
  warnings.warn(message, mplDeprecation, stacklevel=1)
(<Figure size 640x480 with 1 Axes>, [<matplotlib.axes._subplots.AxesSubplot object at 0x1c1133a3c8>])
>>> plt.plot()
```
In case of surfaces, the result is an empty drawing, because the second call creates a second axes on top of the previous one, but draws on the first ax.
```
>>> a = fda.datasets.make_multimodal_samples(ndim_domain=2)
>>> a.plot()
(<Figure size 640x480 with 1 Axes>, [<matplotlib.axes._subplots.Axes3DSubplot object at 0x1c22094400>])
>>> a.plot()
(<Figure size 640x480 with 2 Axes>, [<matplotlib.axes._subplots.Axes3DSubplot object at 0x1c22094400>, <matplotlib.axes._subplots.Axes3DSubplot object at 0x1c22106400>])
>>> plt.show()
```

PR #56, opened 2019-02-26
Plot function of FDataGrid moved to FData

pablomm:
Main changes in the commit:
* Plot function of FDataGrid moved to FData.
* Plot functions of FDataGrid and FDataBasis removed, now are used the common version in FData.
* Now evaluate is called to get the points to plot, this is the unique change made on the @amandaher 
version.
* Temporal plot functions in the examples were removed.

I have two questions about the design of the function:
* Ax and fig arguments can't be passed at the same time, maybe they could be the same argument and parse the type in the function.
* If Ax is passed the return of the function is (None, ax), maybe in this case only the ax should be returned.

I found a bug, documented in #55.


vnmabus:
> * Ax and fig arguments can't be passed at the same time, maybe they could be the same argument and parse the type in the function.

The only downside is that the name of the parameter would be harder to remember. One possible solution is to allow the use of `fig` and `ax` as alias.

> * If Ax is passed the return of the function is (None, ax), maybe in this case only the ax should be returned.

A better solution is to return the figure the axe belong to, with `ax.get_figure()`.

pablomm:
I create a new argument `chart` which could accept figure, ax or a list of axes, and allowing the use of `fig` and `ax` as alias.
https://github.com/GAA-UAM/fda/blob/9d8aef5320be93c1fb09397f63e4a62983f23a0a/fda/functional_data.py#L663

* The new argument allows axes, figures and list of axes 
> fd.plot()
> fd.plot(ax)
> fd.plot([ax])
> fd.plot([ax1, ax2])
> fd.plot(fig)

* The old named arguments could be used too, but I don't know if these alias are useful.
> fd.plot(fig=fig)
> fd.plot(ax=[ax])

* Now is returned the figure of the ax instead of None.
> A better solution is to return the figure the axe belong to, with ax.get_figure().

pablomm:
Bug #55 fixed. Now when the figure it is not passed if the figure geometry (rows, cols) is the same than the current figure the same axes are used, otherwise it is created a new figure. 

Four examples with the behaviour if neither the figured nor the ax is passed.

Compatible plot:
> curve1_1d.plot() # The current fig is empty, will use the current fig adding a new ax
> curve2_1d.plot() # Will be plotted in the same ax than curve 1
> plt.show() # Will clear the current fig and show 1 figure

Different number of dimensions:
> curve1_1d.plot() # The current fig is empty, will use the current fig adding a new ax
> curve_3d.plot() # Last plot has 1 row and 1 column with 1 ax, this new plot has 2 rows, 2 cols with 
                              # 3 axes, so a new figure will be created.
> plt.show() # 2 figures will be shown

Different number of rows or columns:
> curve1_3d.plot(nrows=1, ncols=3) # The current fig is empty, will use the current fig adding a new ax
> curve_3d.plot(nrows=3, ncols=1) # A new figure will be created with 3 rows and 1 col
> plt.show() # 2 figures will be shown

Different type of plot (3d and rectilinear)
> curve1_1d.plot() # Plot of unidimensional function
> surface.plot() # Plot of surface, the projection is 3d, different than the current plot. A new figure
                            # will be created.
> plt.show() # 2 figures will be shown

pablomm:
Closes #53.

fda/functional_data.py
```diff
@@ -516,21 +522,276 @@ def shift(self, shifts, *, restrict_domain=False, extrapolation=None,
...
+
+            for i in range(self.ndim_image):
+                for j in range(self.nsamples):
+                    ax[i].plot_surface(X, Y, Z[j,...,i], **kwargs)
+
+        self.set_labels(fig, ax)
+
+        return fig, ax
 
     def _set_labels(self, ax):
```
amandaher: Is this method still necessary?

fda/functional_data.py
```diff
@@ -516,21 +522,276 @@ def shift(self, shifts, *, restrict_domain=False, extrapolation=None,
...
+        """Set figure and its axes.
 
         Args:
-            ax (axis object, optional): axis over with the graphs are plotted.
-                Defaults to matplotlib current axis.
-            **kwargs: keyword arguments to be passed to the
-                matplotlib.pyplot.plot function.
+            nrows(int, optional): designates the number of rows of the figure to
+                plot the different dimensions of the image. ncols must be also
+                be customized in the same call.
```
amandaher: Could you remove `ncols must be also  be customized in the same call.`? It do not know what I was thinking about when I added it!

fda/functional_data.py
```diff
@@ -516,21 +522,276 @@ def shift(self, shifts, *, restrict_domain=False, extrapolation=None,
...
+            raise ValueError("The number of columns and the number of rows "
+                             "specified is incorrect.")
+
+        if fig is None and ax is None:
+            fig, ax = self.set_figure_and_axes(nrows, ncols)
+
+        if fig is not None:
+            ax = fig.get_axes()
+
+        else:
```
amandaher: `elif ax is not None`?

fda/functional_data.py
```diff
@@ -516,21 +522,276 @@ def shift(self, shifts, *, restrict_domain=False, extrapolation=None,
...
+
+    def set_labels(self, fig=None, ax=None):
+        """Set labels if any.
+
+        Args:
+            fig (figure object): figure object containing the axes that
+                implement set_xlabel and set_ylabel, and set_zlabel in case
+                of a 3d projection.
+
+        """
```
amandaher: This piece of code I think it can be changed to:
¨¨¨
if self.dataset_label is not None:
   fig.suptitle(self.dataset_label)
¨¨¨
I had to do this checks because I did not want to put the same title to all the subplots of the same figure in case the `fig` parameter was None, but now is not going to be None due to the modification in the returning objects of `generic_potting_checks`.

amandaher:
I wrote a few comments but you can merge whenever yo consider.
Good seen the bug, I only tested the function with unidimensional plots and the results were good (the warning was not shown in my Jupyter Notebook), when in fact it should give also an empty plot.

fda/functional_data.py
pablomm: Deleted in ee928f8.

fda/functional_data.py
pablomm: Done.

fda/functional_data.py
pablomm: Changed in ee928f8.

PR #61, opened 2019-03-09
Times function por fdatabasis multiplication

manso92:


fda/basis.py
```diff
@@ -1991,6 +1991,52 @@ def copy(self, *, basis=None, coefficients=None, dataset_label=None,
...
+            (FDataBasis): FDataBasis object containing the multiplication
+
+        """
+        if isinstance(other, FDataBasis):
+
+            if not _same_domain(self.domain_range, other.domain_range):
+                raise ValueError("The functions domains are different.")
+
+            basisobj = self.basis.basis_of_product(other.basis)
+            neval = max(10 * max(self.nbasis, other.nbasis) + 1, 201)
```
vnmabus: I do not want magical numbers in code. If it is really neccessary, it should be a constant.

fda/basis.py
```diff
@@ -1991,6 +1991,52 @@ def copy(self, *, basis=None, coefficients=None, dataset_label=None,
...
+            (left, right) = self.domain_range[0]
+            evalarg = numpy.linspace(left, right, neval)
+
+            first = self.copy(coefficients=(numpy.repeat(self.coefficients,
```
vnmabus: What happens if both have only one sample? Why is that an special case?

manso92: If both has only one sample it will do well. Because this condition only looks if there is a fdatabasis that only has one sample it copies it to the samples of the other object. If both has only one sample, it will do nothing.

fda/basis.py
manso92: This file has other references to numbers like this. Should I make a function constan or a class constant?

vnmabus: File constant, I think.

fda/basis.py
```diff
@@ -1954,6 +1955,53 @@ def copy(self, *, basis=None, coefficients=None, dataset_label=None,
                           axes_labels=axes_labels, extrapolation=extrapolation,
                           keepdims=keepdims)
 
+    def times(self, other):
```
vnmabus: I prefer another name, such as product. Times can be misleading.

PR #64, opened 2019-03-18
Extend functionality on inner product

manso92:
Some changes to make inner product more powerfull


fda/basis.py
```diff
@@ -19,6 +19,7 @@
 
 from . import grid
 from .functional_data import FData, _list_of_arrays
+from .lfd import Lfd
```
vnmabus: Although we have got used to the `Lfd` name, maybe you should try to coin a better class and module name for a future pull request. I should have mentioned it when you first added that class.

fda/basis.py
```diff
@@ -2034,18 +2035,63 @@ def times(self, other):
...
         if isinstance(other, Basis):
             other = other.to_basis()
 
-        inprodmat = [
-            scipy.integrate.quad(lambda x: self[i].evaluate([x]) * other[j].evaluate([x]),
-                                 self.domain_range[0][0], self.domain_range[0][1])[0]
-            for j in range(0, other.nsamples)
-            for i in range(0, self.nsamples)]
-        return numpy.array(inprodmat).reshape((self.nsamples, other.nsamples))
+        # TODO this will be used when lfd evaluation is ready
```
vnmabus: I hope you will address it in the future.

PR #65, opened 2019-03-23
Operations for fdatabasis objects

manso92:
A first approach to fdatabasis operations.
Operations are only between same basis or numbers.


fda/basis.py
```diff
@@ -2189,38 +2190,51 @@ def __getitem__(self, key):
 
         if isinstance(key, int):
             return self.copy(coefficients=self.coefficients[key:key + 1])
-
         else:
             return self.copy(coefficients=self.coefficients[key])
 
+    def plus_samples(self):
```
vnmabus: The name is not clear.

fda/basis.py
```diff
@@ -2189,38 +2190,51 @@ def __getitem__(self, key):
...
+        if isinstance(other, FDataBasis):
+            if self.basis != other.basis:
+                raise NotImplementedError
+            else:
+                return FDataBasis(self.basis.copy(),
+                                  self.coefficients + other.coefficients)
 
-        raise NotImplementedError
+        coefs = self.coefficients.copy()
+        coefs[:, 0] = self.coefficients[:, 0] + numpy.array(other)
```
vnmabus: We should check if `other` can be used as an `ndarray` and return `NotImplemented` otherwise.

fda/basis.py
```diff
@@ -2189,38 +2190,51 @@ def __getitem__(self, key):
...
-        raise NotImplementedError
+        return (-1 * self).__add__(other)
 
     def __mul__(self, other):
         """Multiplication for FDataBasis object."""
+        if isinstance(other, FDataBasis):
+            raise NotImplementedError
 
-        raise NotImplementedError
+        mult = numpy.atleast_2d(other).reshape(-1, 1)
```
vnmabus: Why the reshape?

manso92: Because a need a column vector to apply only on the first coordinate.

fda/basis.py
manso92: How can I check if an object is numpy arrayable?

vnmabus: Try the conversion and catch the possible exception.

skfda/basis.py
```diff
@@ -2208,7 +2208,10 @@ def __add__(self, other):
                                   self.coefficients + other.coefficients)
 
         coefs = self.coefficients.copy()
-        coefs[:, 0] = self.coefficients[:, 0] + numpy.array(other)
+        try:
+            coefs[:, 0] = self.coefficients[:, 0] + numpy.array(other)
+        except:
```
vnmabus: Catch ONLY the exception that is thrown in the conversion.

manso92: We talk that the conversion error will be thrown not caught.

The add mismatch will be treated as an exception

manso92:
Sounds really good. Working on it.

skfda/representation/basis.py
```diff
@@ -324,6 +324,42 @@ def _to_R(self):
...
+            coefs[:, 0] = coefs[:, 0] + other
+        except TypeError:
+            raise NotImplementedError
+
+        return self.copy(), coefs
+
+    def sub_same_basis(self, coefs1, coefs2):
+        return self.copy(), coefs1 - coefs2
+
+    def sub_costant(self, coefs, other):
```
vnmabus: Typo.

skfda/representation/basis.py
```diff
@@ -324,6 +324,42 @@ def _to_R(self):
     def inner_product(self, other):
         return numpy.transpose(other.inner_product(self.to_basis()))
 
+    def add_same_basis(self, coefs1, coefs2):
```
vnmabus: I think those methods should be private. There is no reason for users to mess with them.

skfda/representation/basis.py
```diff
@@ -324,6 +324,42 @@ def _to_R(self):
...
+        return self.copy(), coefs1 + coefs2
+
+    def add_costant(self, coefs, other):
+        coefs = coefs.copy()
+        other = numpy.array(other)
+        try:
+            coefs[:, 0] = coefs[:, 0] + other
+        except TypeError:
+            raise NotImplementedError
```
vnmabus: This should be converted to `return NotImplemented` in the FDataBasis operations. For efficiency, maybe you should return `NotImplemented` here to.

skfda/representation/basis.py
```diff
@@ -2196,22 +2244,32 @@ def __radd__(self, other):
...
     def __rsub__(self, other):
         """Right subtraction for FDataBasis object."""
-
-        raise NotImplementedError
+        return (self * -1).__add__(other)
 
     def __mul__(self, other):
         """Multiplication for FDataBasis object."""
+        if isinstance(other, FDataBasis):
+            raise NotImplementedError
```
vnmabus:
```suggestion
            return NotImplemented
```

skfda/representation/basis.py
```diff
@@ -2196,22 +2244,32 @@ def __radd__(self, other):
...
+        basis, coefs = self.basis.mul_costant(self.coefficients, other)
+
+        return self.copy(basis=basis, coefficients=coefs)
 
     def __rmul__(self, other):
         """Multiplication for FDataBasis object."""
-
         return self.__mul__(other)
 
     def __truediv__(self, other):
```
vnmabus: Should we allow division by constants too?

skfda/representation/basis.py
```diff
@@ -329,14 +329,14 @@ def _to_R(self):
     def inner_product(self, other):
         return numpy.transpose(other.inner_product(self.to_basis()))
 
-    def add_same_basis(self, coefs1, coefs2):
+    def _add_same_basis(self, coefs1, coefs2):
         return self.copy(), coefs1 + coefs2
 
-    def add_costant(self, coefs, other):
+    def _add_costant(self, coefs, constant):
```
vnmabus: Typo

skfda/representation/basis.py
vnmabus: This has not been addressed.

manso92:
'''
def sum_samples(self):	
        """Sums all the samples on the object to a single sample object"""	
        return self.copy(coefficients=numpy.sum(self.coefficients, axis=0))
'''

manso92:
This method is no longer necessary 
```python
    def sum_samples(self):	
        """Sums all the samples on the object to a single sample object"""	
        return self.copy(coefficients=numpy.sum(self.coefficients, axis=0))	

```
instead of using

```python
    fd = FDataBasis(...)
    fd.sum_samples()
```

you should use

```python
    fd = FDataBasis(...)
    sum(fd)
```

PR #68, opened 2019-04-07
Added len and iter methods

pablomm:
Proposal to allow iteration on the samples of the fdata objects

```python
for sample in fd:
   do_something(sample)
```

and 

```python
for i in range(len(fd)):
   do_something(fd[i])
```
This will allow unzip the objects too:
```python
a, b = fda.datasets.make_gaussian_process(n_samples=2)
```
Or with the * operator
```python
do_something(*fd)
```
With these changes some interactions with other objects will change
```python
fd = fda.datasets.make_gaussian_process(n_samples=2)
len([fd]) # Will return 1
len(list(fd)) # Now will return 2 
```


fda/functional_data.py
```diff
@@ -1011,4 +1011,15 @@ def __truediv__(self, other):
     def __rtruediv__(self, other):
         """Right division for FData object."""
 
-        pass
\ No newline at end of file
+        pass
+
+    def __iter__(self):
```
vnmabus: I though that we already had the iteration, as we implement `__getitem__`. Is it not true?

pablomm: Yes, it could be iterated, but the result is an infinite loop.
```python
for sample in fd: # fd without __iter__
   do_something(sample) # Infinite loop
```
And the objects cannot be unzipped.

Without `__len__` the iteration works well, but I like the syntax `len(fd)`.

PR #73, opened 2019-04-08
Fixed problem with colors

### LINKED ISSUE #72: Problem with colors when are rgb tuples
The following example crashed due to the use of numpy array to store the colors added in #66
https://github.com/GAA-UAM/fda/blob/940caa755e40cf26395e22a772806cba3ff18fd7/examples/plot_pairwise_alignment.py#L126-L133

pablomm:
Closes #72.
Lists used to store the colors instead of numpy arrays.

### DIFF
--- a/fda/functional_data.py
+++ b/fda/functional_data.py
@@ -930,13 +930,11 @@ class FData(ABC):
         else:
 
             if 'color' in kwargs:
-                sample_colors = np.asarray(
-                    kwargs.get("color")).repeat(self.nsamples)
+                sample_colors = self.nsamples * [kwargs.get("color")]
                 kwargs.pop('color')
 
             elif 'c' in kwargs:
-                sample_colors = np.asarray(
-                    kwargs.get("c")).repeat(self.nsamples)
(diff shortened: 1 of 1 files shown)


fda/functional_data.py
```diff
@@ -930,13 +930,11 @@ def plot(self, chart=None, *, derivative=0, fig=None, ax=None, nrows=None,
...
+                sample_colors = self.nsamples * [kwargs.get("color")]
                 kwargs.pop('color')
 
             elif 'c' in kwargs:
-                sample_colors = np.asarray(
-                    kwargs.get("c")).repeat(self.nsamples)
+                sample_colors = self.nsamples * [kwargs.get("color")]
                 kwargs.pop('c')
 
             else:
```
vnmabus: What about the else?

pablomm: I only made the minimal changes to fix the bug.
As far I know the else won't crash for the type of color. I think that this block of code should be rewrite using a generator and maybe moved to a plotting module.

Issue #77, opened 2019-04-13
Random seed in tests


pablomm:
The random state of the tests of magnitude shape plot is not being fixed, which makes the tests fail.

https://github.com/GAA-UAM/fda/blob/28a6e0b58053c29dd4fca0624d85ee44e00bbf19/fda/magnitude_shape_plot.py#L378

In the example of surface boxplot the random seed isn't being fixed either to generate the datasets.

vnmabus:
Thank you for finding the bug, Pablo!

@amandaher, you should expose a random seed if there is randomness in a function. All seeds must be fixed in tests/examples to have reproducibility. You can look at the dataset generation utilities to see an example.

vnmabus:
In fact, in this case, if the randomness come from the covariance estimation, I would say that the least surprising behaviour is to have a fixed seed by default, but allow changing it. What do you think?

PR #79, opened 2019-04-15
random_state

amandaher:
Solving issue #77.

### DIFF
--- a/examples/plot_surface_boxplot.py
+++ b/examples/plot_surface_boxplot.py
@@ -34,9 +34,11 @@ from fda.datasets import make_sinusoidal_process, make_gaussian_process
 n_samples = 10
 n_features = 10
 
-fd1 = make_sinusoidal_process(n_samples = n_samples, n_features=n_features)
+fd1 = make_sinusoidal_process(n_samples = n_samples, n_features=n_features,
+                              random_state=5)
 fd1.dataset_label = "Sinusoidal process"
-fd2 = make_gaussian_process(n_samples = n_samples, n_features=n_features)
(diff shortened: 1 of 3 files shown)


amandaher:
@pablomm I am following the changes you suggested in this pull request, however, I do not understand why this should solve the failing of the tests.

pablomm:
> @pablomm I am following the changes you suggested in this pull request, however, I do not understand why this should solve the failing of the tests.

As @vnmabus says in #77, `MinCovDet` uses some type of random estimation of the variance, and the result of the magnitude shape plot won't have to be the same in all the executions, which is why tests sometimes fail. 

Setting the random state the result of the test will always be the same.

amandaher:
I understand that, however, since the error is the below one:
`(mismatch 8.57142857142857%)
x: array([0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 0, 1, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 0, 1, 0, 1, 0, 1, 1, 1]) 
y: array([0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 0, 1, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 0, 1, 0, 0, 0, 0, 0, 1])`,
it seems to me a problem related to the precision of the comparison rather than to the randomness. 
Nevertheless,  giving the  possibility of specifying the random state is an upgrade for the code.

pablomm:
But the last elements of the arrays are different, it is not for the tolerance.
```
x: array([..., 1, 0, 1, 1, 1]) 
y: array([..., 0, 0, 0, 0, 1])
```

PR #80, opened 2019-04-19
Feature/fix cython build

### LINKED ISSUE #78: ImportError running test cases
Hi developer,

Sorry if my question is more like seeking help rather than contributing to the package development. I am trying to run your fda package locally while I could not build it successfully by running into:
**ImportError: dynamic module does not define module export function (PyInit_optimum_reparam_extension)**,  and error: 
**import optimum_reparam_extension**.
I wish this package performs similar function as the fda package in R. I am brand new to Github, so all I did after downloading the source code from Github using "git" command is installing all the requirements in requirements.txt and building the project by running "**python3 setup.py <operations>**". For <operations>, I ran "build", "install", "bdist" and "check". If it ever helps, I am using Python 3.6.4 on a Mac Mojave 10.14. It would be more than great to get your help on where might be going wrong. I indeed have no idea if I should expect this "optimum_reparam_extension" file to be originally in your package, or it should be something generated after I build it myself. Great many thanks!

pablomm:
Changes to fix problem with Cython in some osX distributions. Closes #78.

### DIFF
--- a/deps/fdasrsf/optimum_reparam.pyx
+++ b/deps/fdasrsf/optimum_reparam.pyx
@@ -1,4 +1,5 @@
-# cython: language_level=3
+#!python
+#cython: language_level=3
 
 
 # Original code developed by J. Derek Tucker.
@@ -52,7 +53,7 @@ def coptimum_reparam_n(np.ndarray[double, ndim=1, mode="c"] mq,
     sizes = np.zeros(N, dtype=np.int32)
     Go = np.zeros((M, N))
     To = np.zeros((M, N))
-    for k in xrange(0, N):
+    for k in range(0, N):
         qi = q[:, k] / norm(q[:, k])
         qi = np.ascontiguousarray(qi)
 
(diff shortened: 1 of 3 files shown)


deps/fdasrsf/optimum_reparam.pyx
```diff
@@ -1,4 +1,5 @@
-# cython: language_level=3
+#!python
```
vnmabus: Why the shebang? And why Python?

deps/fdasrsf/optimum_reparam.pyx
```diff
@@ -52,7 +53,7 @@ def coptimum_reparam_n(np.ndarray[double, ndim=1, mode="c"] mq,
     sizes = np.zeros(N, dtype=np.int32)
     Go = np.zeros((M, N))
     To = np.zeros((M, N))
-    for k in xrange(0, N):
```
vnmabus: How did that work for you, when you changed the language level to 3?

pablomm: I suppose that xrange it is supported for cython in python 3. I changed it to be consistent with the python3 style (It's just speculation).

deps/fdasrsf/optimum_reparam.pyx
pablomm: Looking for solutions to #78 I saw a cython code with the same problem and the code fixed had this shebang.
The problem was because the extension and the .pyx file had a different name, doesn't change anything, but I preferred not to change it just in case.

PR #85, opened 2019-04-23
Cython as dependency

### LINKED ISSUE #83: Add Cython as a dependency
Cython should be a dependency. [Here](https://stackoverflow.com/questions/18023546/marking-cython-as-a-build-dependency) says how to do that.

pablomm:
Closes #83.


pablomm:
@vnmabus, the cython dependency should be in `requirements.txt`?

https://github.com/GAA-UAM/fda/blob/749c8b89a7801f854fcd5e141351ea9767237a25/requirements.txt#L5

vnmabus:
Closed as #86 solves the same problem.

PR #93, opened 2019-05-03
clustering functionality

amandaher:
Algorithms Kmeans and FuzzyKMeans implemented in two classes  named as written with the same paren class, ClusteringData.


skfda/clustering.py
```diff
@@ -0,0 +1,1180 @@
...
+import matplotlib.patches as mpatches
+from matplotlib.ticker import MaxNLocator
+import random
+from abc import ABC, abstractmethod
+
+__author__ = "Amanda Hernando Bernabé"
+__email__ = "amanda.hernando@estudiante.uam.es"
+
+
+class ClusteringData(ABC):
```
vnmabus: I would change the name, as this ABC is only for K-means and Fuzzy K-means, and not for other algorithms, such as Hierarchical clustering.

skfda/clustering.py
```diff
@@ -0,0 +1,1180 @@
...
+    @property
+    def fdatagrid(self):
+        return self._fdatagrid
+
+    @property
+    def n_clusters(self):
+        return self._n_clusters
+
+    @property
+    def init(self):
```
vnmabus: This name is very confusing.

skfda/clustering.py
```diff
@@ -0,0 +1,1180 @@
...
+        if max_iter < 1:
+            raise ValueError(
+                "The number of iterations must be greater than 0.")
+
+        self._max_iter = max_iter
+        self._random_state = random_state
+        self._p = p
+
+    @property
+    def fdatagrid(self):
```
vnmabus: I would not store the data.

skfda/clustering.py
```diff
@@ -0,0 +1,1180 @@
...
+__email__ = "amanda.hernando@estudiante.uam.es"
+
+
+class ClusteringData(ABC):
+    """Base class to implement clustering algorithms.
+
+    Class from which both :class:`K-Means <fda.clustering.KMeans>` and
+    :class:`Fuzzy K-Means <fda.clustering.FuzzyKMeans>` classes inherit."""
+
+    def __init__(self, max_iter, random_state, p):
```
vnmabus: Please, check all estimators with the utilities explained in https://scikit-learn.org/stable/developers/contributing.html#rolling-your-own-estimator, because I am not sure that you are following all the scikit-learn conventions.

skfda/clustering.py
```diff
@@ -0,0 +1,1180 @@
...
+        """Sets the arguments *max_iter* and *random_state* and *p*.
+
+        Args:
+            max_iter (int): Maximum number of iterations of the clustering
+                algorithm.
+            random_state (int, RandomState instance or None): Defaults to 0.
+                Determines random number generation for centroid initialization.
+                Use an int to make the randomness deterministic. See Glossary.
+                https://scikit-learn.org/stable/glossary.html#term-random-state
+            p (int): Identifies the p-norm used to calculate the distance
```
vnmabus: I would allow an arbitrary metric.

skfda/clustering.py
```diff
@@ -0,0 +1,1180 @@
...
+                "The number of observations must be greater than 1.")
+
+        if n_clusters < 2:
+            raise ValueError(
+                "The number of clusters must be greater than 1.")
+
+        if init is not None and init.shape != (
+                fdatagrid.ndim_image, n_clusters, fdatagrid.ncol):
+            raise ValueError(
+                "The init ndarray should be of shape (ndim_image, n_clusters, n_features) "
```
vnmabus: Why this shape? I would expect that a fdatagrid is received, with as many observations as clusters.

skfda/clustering.py
```diff
@@ -0,0 +1,1180 @@
...
+        """
+        comparison = True
+        random_state_aux = self.random_state
+        while comparison:
+            random.seed(a=random_state_aux)
+            indices = random.sample(range(fdatagrid.nsamples), n_clusters)
+            centers = data_matrix[indices]
+            comparison = np.asarray([(centers[i] == centers).all()
+                                     for i in range(n_clusters)]).sum()
+            random_state_aux += 1
```
vnmabus: Can this be done for RandomState objects?

skfda/clustering.py
```diff
@@ -0,0 +1,1180 @@
...
+            n_clusters (int): Number of groups into which the samples are
+                classified.
+
+        Returns:
+            centers (ndarray): initial centers
+        """
+        comparison = True
+        random_state_aux = self.random_state
+        while comparison:
+            random.seed(a=random_state_aux)
```
vnmabus: Do NOT change global state in a library.

skfda/clustering.py
```diff
@@ -0,0 +1,1180 @@
...
+            indices = random.sample(range(fdatagrid.nsamples), n_clusters)
+            centers = data_matrix[indices]
+            comparison = np.asarray([(centers[i] == centers).all()
```
vnmabus: What are you trying to do?

skfda/clustering.py
amandaher: The idea is that if other algorithms were implemented, they could inherit from this class too.

skfda/clustering.py
amandaher: It is the name used in sklearn Kmeans class.

skfda/clustering.py
amandaher: An array with as many observations as clusters is expected, but ndim_image times. There should be the possibility to initialize different centroids for each dimension of the image.
I could change it to expect an fdatagrid, in that case, its data_matrix should be of the shape (n_clusters, n_features, ndim_image).

skfda/clustering.py
amandaher: I am using the random module to calculate the initial centroids randomly. If I used numpy random module would it be fine? The code would be:
```
    def _random_initialization_centers(self, data_matrix, fdatagrid,  n_clusters):
        comparison = True
        random_state_aux = self.random_state
        while comparison:
            np.random.seed(a=random_state_aux)
            indices = np.random.choice(fdatagrid.nsamples, n_clusters, replace=False)
            centers = data_matrix[indices]
            comparison = np.asarray([(centers[i] == centers).all()
                                     for i in range(n_clusters)]).sum()
            random_state_aux += 1

        return centers
```
If we want to obtain the same centroids, that is, reproduce the same result, the random seed must be customized since the numpy.random functions do not accept the seed as a parameter.

skfda/clustering.py
amandaher: I was trying to see that the centers were unique, thus, there were not repeated centroids but it is wrong.
Changed.

skfda/clustering.py
vnmabus: But not all clustering algorithms need those arguments.

skfda/clustering.py
vnmabus: In this case, I guess it is ok, for compatibility.

skfda/clustering.py
vnmabus: I think a FDataGrid is better. Maybe you can allow an automatic conversion from an array, but the shape should be the same as the FDataGrid.

skfda/clustering.py
amandaher: If the new code is the proposed in the below question, the parameter of the class would only accept random_seed, which in that case, it is an Int.

skfda/clustering.py
vnmabus: No, the key is making a RandomState object, which has the necessary methods (https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.RandomState.html). You can use as an example the code for the synthetic dataset generation: https://github.com/GAA-UAM/scikit-fda/blob/a7f32d156468b5ff21f70565be5fd9bcbab68ee2/skfda/datasets/_samples_generators.py#L35-L54

amandaher: https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_random_state.html It says that it returns a new RandomState that it is what I am doing.

vnmabus: The user should be able to provide a seed, a RandomState object or None. That is why we use `sklearn.utils.check_random_state`.

skfda/clustering.py
pablomm: Should `ClusteringData`  inherit from the sklearn classes `BaseEstimator`, `ClusterMixin` and `TransformerMixin`?

https://github.com/scikit-learn/scikit-learn/blob/7b136e92acf49d46251479b75c88cba632de1937/sklearn/cluster/k_means_.py#L771
https://github.com/scikit-learn/scikit-learn/blob/7b136e92acf49d46251479b75c88cba632de1937/sklearn/base.py

vnmabus: I think so.

skfda/clustering.py
amandaher: I have used `check_estimator` to see if the implemented estimators `KMeans` and `FuzzyKMeans` pass the test. It fails in both cases because the first parameter of the `fit` function expects a numpy.ndarray instead of an FDataGrid object., and in my opinion it should be like this.

PR #99, opened 2019-05-14
Derivative

manso92:
Exact derivative for fdatabasis objects


skfda/representation/basis.py
```diff
@@ -331,7 +335,9 @@ def __repr__(self):
 
     def __eq__(self, other):
         """Equality of Basis"""
-        return type(self) == type(other) and _same_domain(self.domain_range, other.domain_range) and self.nbasis == other.nbasis
+        return (type(self) == type(other)
```
vnmabus: This will have to be overridden when you implement the custom basis for FPCA, right?

skfda/representation/basis.py
```diff
@@ -374,6 +380,10 @@ def _ndegenerated(self, penalty_degree):
         """
         return penalty_degree
 
+    def _derivative(self, coefs, order=1):
+        return (self.copy(), coefs if order == 0
```
vnmabus: I am not sure if we allow the coefficients to be changed, but if we do, this should be a copy. Maybe even is better that a copy is always passed to this function in that case, to prevent possible mistakes.

skfda/representation/basis.py
```diff
@@ -543,6 +553,11 @@ def _compute_matrix(self, eval_points, derivative=0):
 
         return mat
 
+    def _derivative(self, coefficients, order=1):
```
vnmabus: The name of all parameters should be the same as in the overridden function.

skfda/representation/basis.py
```diff
@@ -1071,7 +1099,26 @@ def _to_R(self):
         return ("create.bspline.basis(rangeval = c(" + str(drange[0]) + "," +
                 str(drange[1]) + "), nbasis = " + str(self.nbasis) +
                 ", norder = " + str(self.order) + ", breaks = " +
-                self._listtoR(self.knots) + ")")
+                self._list_to_R(self.knots) + ")")
+
+    def _to_scipy_BSpline(self, coefs):
+        from scipy.interpolate import BSpline as SciBSpline
```
vnmabus: Why this import is local?

skfda/representation/basis.py
manso92: Yes

PR #100, opened 2019-05-16
Transformer to reduce the image dimension

pablomm:
Created a transformer which receives a multivariate function from R^n->R^d and applies a vectorial
norm to reduce the image dimension to 1.

There are two version of the transformer, a procedural one and a sklearn style transformer.

I put the functions in skfda.preprocessing.dim_reduction, but maybe there is a better place.


pablomm:
It is used in the multivariate version of the Lp norm (in process), I can move it to the `metrics` module.
I though that in cases like the study of the velocity of trajectories (the euclidean norm of the derivative) would be useful as a preprocessing step in a pipeline.

vnmabus:
It seems too specific to be a transformer I think. But the function should be available in metrics.

pablomm:
Do I keep the sklearn like version?

vnmabus:
I do not think it is necessary.

pablomm:
I moved the vectorial norm to `metrics.py` and the `norm_lp` and `lp_distance` now accept functions from R to R^k.

skfda/misc/metrics.py
```diff
@@ -54,12 +55,66 @@ def _cast_to_grid(fdata1, fdata2, eval_points=None):
...
+            and if p is -inf it is applied min(abs(x)). See numpy.linalg.norm
+            to more information. Defaults to 2.
+
+    Returns:
+        (:class:`FDatagrid`): FDatagrid with image dimension equal to 1.
+
+    Examples:
+
+        >>> from skfda.datasets import make_multimodal_samples
+        >>> from skfda.preprocessing.dim_reduction import vectorial_norm
```
pablomm:
```suggestion
        >>> from skfda.misc.metrics import vectorial_norm
```

PR #118, opened 2019-06-05
Gramian matrix for basis and improvements on inner product

manso92:


skfda/representation/basis.py
```diff
@@ -325,6 +325,70 @@ def _list_to_R(self, knots):
     def _to_R(self):
         raise NotImplementedError
 
+    def inner_matrix(self, other=None):
```
vnmabus: Should it be public?

skfda/representation/basis.py
```diff
@@ -325,6 +325,70 @@ def _list_to_R(self, knots):
...
+
+        gramian = np.zeros((self.nbasis, other.nbasis))
+
+        for i in range(self.nbasis):
+            for j in range(other.nbasis):
+                gramian[i, j] = first[i].inner_product(second[j], None, None)
+
+        return gramian
+
+    def gramian_matrix(self):
```
vnmabus: Should it be public?

vnmabus: I prefer `gram_matrix`. It is personal taste, however.

skfda/representation/basis.py
```diff
@@ -2175,6 +2239,13 @@ def inner_product(self, other, lfd_self=None, lfd_other=None,
         if weights is not None:
             other = other.times(weights)
 
+        if self.nsamples * other.nsamples > self.nbasis * other.nbasis:
+            return self.coefficients @ self.basis.inner_matrix(other.basis) @ other.coefficients.T
+        else:
+            return self._inner_product_integrate(other, lfd_self, lfd_other)
+
+    def _inner_product_integrate(self, other, lfd_self, lfd_other):
```
vnmabus: Is this still necessary?

skfda/representation/basis.py
manso92: I think so. The gramian matrix of a basis is an option you can ask for compute

skfda/representation/basis.py
manso92: I think different with this, I don't know how this method could help. I'll put it private.

skfda/representation/basis.py
manso92: Nice

skfda/representation/basis.py
manso92: Yes, because if you want to calculate the inner product between two objects with 1 sample each with 10 basis each, it computes only one integral instead of the 100 integrals it has to make for the inner matrix and then the matrix product. I filter the method base on the number of integrals it will do
```python
if self.nsamples * other.nsamples > self.nbasis * other.nbasis:
    inner matrix
else:
    numerical integration
```

This is the check for computing the minimun number of integrals

vnmabus: Seems reasonable.

vnmabus:
The tests are failing, because you have not updated the name of the function in the tests.

skfda/representation/basis.py
```diff
@@ -326,6 +326,70 @@ def _list_to_R(self, knots):
...
+            numpy.array: Gram Matrix of the basis.
+
+        """
+        fbasis = self.to_basis()
+
+        gram = np.zeros((self.nbasis, self.nbasis))
+
+        for i in range(fbasis.nbasis):
+            for j in range(i, fbasis.nbasis):
+                gram[i, j] = fbasis[i].inner_product(fbasis[j], None, None)
```
pablomm: This integration can be avoided for the most common cases, using an explicit formula, if it's still running slow.
For example , in the polynomial basis in `[a, b]`, 

`gram[i, j] = <t^i, t^j> = int_a^b t ^(i + j) dt = (b^(i+j+1) - a^(i+j+1)) / (i+j+1)`.

Maybe could be added as an issue override this methods when the product could be make without numeric integration.

vnmabus: Add it.

manso92: Yo pensé en abrir el issue una vez terminado como con las operaciones. Hice los casos generales y dejé para futuro los casos particulares.

manso92:
I fix test on a commit seconds after you comment

PR #124, opened 2019-06-23
Some tests waring fixes

manso92:
Changes to fix some test warnings

image

### DIFF
--- a/tests/test_elastic.py
+++ b/tests/test_elastic.py
@@ -2,8 +2,6 @@ import unittest
 
 import numpy as np
 
-import matplotlib.pyplot as plt
-
 from skfda import FDataGrid
 from skfda.datasets import make_multimodal_samples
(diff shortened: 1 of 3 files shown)


vnmabus:
I need more information on the warnings. Do they appear in Windows only? How the proposed solution fixes them?

manso92:
I still haven't been able to pass the tests on windows. This error belongs to the matplotlib library with a connection to gtk. This avoids warning by avoiding the use of gtk.

vnmabus:
So, **WHERE** do you have found the warning? It does not happen in my machine.

manso92:
On ubuntu 18.04 it makes this warning if you use this script

```bash
git clone git@github.com:GAA-UAM/scikit-fda.git
cd scikit-fda
python3 setup.py test 
```

PR #125, opened 2019-06-24
Smoothing refactor

vnmabus:
* Change the smoothing methods so that they have the scikit-learn API an can be used in a pipeline.

### DIFF
--- a/skfda/preprocessing/smoothing/validation.py
+++ b/skfda/preprocessing/smoothing/validation.py
@@ -125,7 +125,8 @@ class SmoothingParameterSearch(GridSearchCV):
         param_values (iterable): iterable containing the values to test
             for *smoothing_parameter*.
         scoring (scoring method): scoring method used to measure the
-            performance of the smoothing.
+            performance of the smoothing. If ``None`` (the default) the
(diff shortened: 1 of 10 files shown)


skfda/preprocessing/smoothing/validation.py
```diff
@@ -54,72 +62,90 @@ def gcv(fdatagrid, s_matrix, penalisation_function=None):
...
+    Note:
+        This is similar to fitting a scikit-learn GridSearchCV over the
+        data, using the cv_method as a scorer.
 
-            }
+    Args:
+        estimator (smoother estimator): scikit-learn compatible smoother.
+        param_values (iterable): iterable containing the values to test
+            for *smoothing_parameter*.
+        scoring (scoring method): scoring method used to measure the
```
pablomm: Only with our documentation I haven't been able to infer the default scorer. 
I had to look at the scikit-learn documentation to see that the estimator score method is used.

vnmabus: Improved in f7f1313b51dbd283d451a205969d6d28c19dc80b.

PR #129, opened 2019-06-27
Feature/covariances

vnmabus:
Add more covariance functions.
Improve the visualization of covariance functions, with several plots.


tests/test_covariances.py
```diff
@@ -0,0 +1,64 @@
+import unittest
+
+import numpy as np
+import skfda
+
+
+class TestsSklearn(unittest.TestCase):
```
manso92: El nombre del test no debería ser TestSklearnCovariances por ser del módulo que prueba

vnmabus: The 'covariance' name already appears in the module name, so it seemed redundant to me.

PR #137, opened 2019-07-11
Delete __author__ and __email__ variables

pablomm:
Deleted the author variables of the modules, as we spoke in one of the last meetings, because now there is the contributors file and most of the modules have been substantially modified by several people.

I have deleted variables from all of you, I put you as reviewers to give your approval.


pablomm:
I close this issue, in case of eliminating the variables it is easier to do another PR than to solve the conflicts when doing update.

PR #139, opened 2019-07-23
Feature/basis smoother

vnmabus:
Move smoothing using basis representation to the smoothing module.


skfda/preprocessing/smoothing/_basis.py
```diff
@@ -0,0 +1,520 @@
...
+        >>> fd_basis.coefficients.round(2)
+        array([[ 0.18,  0.07,  0.09]])
+
+        >>> from skfda.misc import LinearDifferentialOperator
+        >>> fd = skfda.FDataGrid(data_matrix=x, sample_points=t)
+        >>> basis = skfda.representation.basis.Fourier((0, 1), nbasis=3)
+        >>> smoother = skfda.preprocessing.smoothing.BasisSmoother(
+        ...                basis, method='qr',
+        ...                smoothing_parameter=1,
+        ...                penalty=LinearDifferentialOperator(weights=[3, 5]),
```
pablomm: I think the differential operator is not being built properly (or printed). 

```python
>>> smoother.penalty
LinearDifferentialOperator(
    nderiv=1,
    bwtlist=[
    FDataBasis(
        basis=Constant(domain_range=[array([0, 1])], nbasis=1),
        coefficients=[[3]],
        dataset_label=None,
        axes_labels=None,
        extrapolation=None,
        keepdims=False)]
    )
```

skfda/preprocessing/smoothing/_basis.py
```diff
@@ -0,0 +1,520 @@
...
+                 basis,
+                 *,
+                 smoothing_parameter: float = 0,
+                 weights=None,
+                 penalty: Union[int, Iterable[float],
+                                'LinearDifferentialOperator'] = None,
+                 penalty_matrix=None,
+                 output_points=None,
+                 method='cholesky',
+                 keepdims=False,
```
pablomm: I think the ``keepdims`` parameter shouldn't be here. It should be discussed whether to eliminate the keepdims parameter from the ``FData``, leaving it only as an argument for the evaluate function. Anyway, ``keepdims`` should inherit the value of the object to be smoothed.

skfda/preprocessing/smoothing/_basis.py
```diff
@@ -0,0 +1,520 @@
...
+
+    def __init__(self,
+                 basis,
+                 *,
+                 smoothing_parameter: float = 0,
+                 weights=None,
+                 penalty: Union[int, Iterable[float],
+                                'LinearDifferentialOperator'] = None,
+                 penalty_matrix=None,
+                 output_points=None,
```
pablomm: This parameter ``output_points`` is not in the documentation.

Apart from that, I like this parameter name, there are a lot of functions with a similar parameter but the name `eval_points`, in my opinion `output_points` is a better name.

skfda/preprocessing/smoothing/_basis.py
```diff
@@ -0,0 +1,520 @@
...
+                             f"({data_matrix.shape[0]}).")
+
+        fdatabasis = FDataBasis(
+            basis=self.basis, coefficients=coefficients,
+            keepdims=self.keepdims)
+
+        if self.return_basis:
+            return fdatabasis
+        else:
+            return fdatabasis(self.output_points_)
```
pablomm: According to the documentation, an FDataGrid should be returned in this case.
```suggestion
            return fdatabasis.to_grid(eval_points=self.output_points_)
```

skfda/representation/basis.py
```diff
@@ -1589,38 +1592,33 @@ def __init__(self, basis, coefficients, *, dataset_label=None,
         super().__init__(extrapolation, dataset_label, axes_labels, keepdims)
 
     @classmethod
-    def from_data(cls, data_matrix, sample_points, basis, weight_matrix=None,
-                  smoothness_parameter=0, penalty_degree=None,
-                  penalty_coefficients=None, penalty_matrix=None,
+    def from_data(cls, data_matrix, sample_points, basis,
```
pablomm: Why not keep all possible parameters in the ``BasisSmoother`` class?

skfda/preprocessing/smoothing/__init__.py
```diff
@@ -1,2 +1,3 @@
 from . import kernel_smoothers
 from . import validation
+from ._basis import BasisSmoother
```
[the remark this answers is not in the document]
vnmabus: I though about this, but I was not sure. The other smoothers are all kernel smoothers, so a submodule made sense, but in this case, it did not.

skfda/preprocessing/smoothing/_basis.py
vnmabus: You are right. I should also add a test for this behaviour.

skfda/representation/basis.py
vnmabus: Because it is not the same, conceptually, to change the representation than to smooth the data. It was also Alberto's opinion.

pablomm: Okay then. Although the conversion to bases already performs a smoothing. I would consider allowing arguments to pass through ``**kwargs``.

skfda/preprocessing/smoothing/__init__.py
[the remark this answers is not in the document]
pablomm: Having only 4 smoothers, which receive an ``FDataGrid`` and return another ``FDataGrid`` by default, I don't know if a submodule is necessary. In my opinion, it is more important that related objects, with the same function and structure, have the same hierarchy in the api.

I would put them all in the ``smoothing`` module directly, because the functionality of the objects has more weight for me, but it is true that theoretically are two different types of smoothers. I guess it's a personal point of view.

skfda/preprocessing/smoothing/_basis.py
vnmabus: Added to the docs in 73ec1d07ea2d6aae469481a3ed30eaa426f4b273. I am not opposed to the change of name in other places.

skfda/preprocessing/smoothing/_basis.py
vnmabus: Changed, and added case in docstring in 73ec1d07ea2d6aae469481a3ed30eaa426f4b273.

skfda/preprocessing/smoothing/_basis.py
vnmabus: `keepdims` usage eliminated in this module in 73ec1d07ea2d6aae469481a3ed30eaa426f4b273.

skfda/preprocessing/smoothing/_basis.py
vnmabus: It was the `__repr__` method. Changed also in 73ec1d07ea2d6aae469481a3ed30eaa426f4b273.

skfda/preprocessing/smoothing/__init__.py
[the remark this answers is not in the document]
vnmabus: I will not change them for now, because they are related, and deserve an specific submodule. For now, I think that aliasing them directly in the `smoothing` module could be confusing, but I could reconsider that when more smoothers are added.

skfda/representation/basis.py
vnmabus: I know that there is an implicit smoothing, but I also agree with Alberto in that, if you want to add a penalization, it is more clean to use the smoother.

pablomm:
I see that this PR is blocked by the ``keepdims`` parameter removed in 73ec1d0, maybe it is better to leave it as it was and if it is changed do it in another pull request.

vnmabus:
> I see that this PR is blocked by the `keepdims` parameter removed in [73ec1d0](https://github.com/GAA-UAM/scikit-fda/commit/73ec1d07ea2d6aae469481a3ed30eaa426f4b273), maybe it is better to leave it as it was and if it is changed do it in another pull request.

I forgot to check the test and I though it worked. It works now, but I agree that we should remove the keepdims parameter, or change how it works.

PR #148, opened 2019-08-22
Feature/outlier detection

vnmabus:
First version of outlier detectors.

The parameters c and m of the DirectionalOutlierDetector must be computed in a better way, using the asympthotic expression and the simulations, but in a future PR.


skfda/exploratory/outliers/_directional_outlyingness.py
```diff
@@ -0,0 +1,371 @@
...
+    \tilde{Y}}^*_J\right) \left( \mathbf{Y}_i - \mathbf{\tilde{Y}}^*_J
+    \right)^T`. The sub-sample of size h controls the robustness of the method.
+
+    Then, the tail of this distance distribution is approximated as follows:
+
+    .. math::
+        \frac{c\left(m - p\right)}{m\left(p + 1\right)}RMD^2\left(
+        \mathbf{Y}, \mathbf{\tilde{Y}}^*_J\right)\sim F_{p+1, m-p}
+
+    where :math:`p` is the dmension of the image, and :math:`c` and :math:`m`
```
pablomm:
```suggestion
    where :math:`p` is the dimension of the image, and :math:`c` and :math:`m`
```

PR #150, opened 2019-08-28
Feature/fix magnitude shape

vnmabus:
Fix errors in the Magnitude Shape plot, and add an example of the different kinds of outliers.


skfda/exploratory/depth/multivariate.py
```diff
@@ -0,0 +1,33 @@
...
+
+    if pointwise is False:
+        raise NotImplementedError("Only implemented pointwise")
+
+    if X.ndim_codomain == 1:
+        # Special case, can be computed exactly
+        m = X.data_matrix[..., 0]
+
+        return (np.abs(m - np.median(m, axis=0)) /
+                scipy.stats.median_absolute_deviation(m, axis=0))
```
pablomm: I have the following error while running the example ``examples/plot_magnitude_shape_synthetic.py``:

> AttributeError: module 'scipy.stats' has no attribute 'median_absolute_deviation'

I have the scipy version ``1.1.0`` and it was added in the version ``1.3.0``, we should add to the requirements and maybe increase the coverage.

After update scipy all runs ok.

skfda/exploratory/outliers/_directional_outlyingness.py
```diff
@@ -2,15 +2,16 @@
...
+from scipy.stats import f
+import scipy.stats
 from sklearn.base import BaseEstimator, OutlierMixin
 from sklearn.covariance import MinCovDet
 
 import numpy as np
-from skfda.exploratory.depth import modified_band_depth
+from skfda.exploratory.depth.multivariate import projection_depth
 
+from . import _directional_outlyingness_experiment_results as experiments
```
pablomm: The values of the experiments are only used in ``_parameters_numeric ``, maybe this import could be moved to the function, in order to avoid load the module with the data.

skfda/exploratory/depth/__init__.py
```diff
@@ -0,0 +1,4 @@
+from ._depth import (band_depth,
+                     modified_band_depth,
+                     fraiman_muniz_depth,
+                     outlyingness_to_depth)
```
pablomm: ``projection_depth`` was excluded from this import for any reason?

skfda/exploratory/outliers/_directional_outlyingness_experiment_results.py
```diff
@@ -0,0 +1,93 @@
+# flake8: noqa
```
pablomm: A comment with the reference of the source of this data or the method to generate it would be useful.

skfda/exploratory/depth/multivariate.py
vnmabus: Fixed in 4c53ffa658485efce3bb121c6df1fbb9640bc46b.

skfda/exploratory/outliers/_directional_outlyingness.py
vnmabus: Fixed in 25a09c57a3a259d4a4df9506b6f4af06e4d24ad5.

skfda/exploratory/depth/__init__.py
vnmabus: `projection_depth` is not added directly as it is not a functional depth, but the submodule for multivariate depths and the documentation are added in 0fafea0fb1b22a6167a23f446dd9b345d16e2568.

skfda/exploratory/outliers/_directional_outlyingness_experiment_results.py
vnmabus: Added in 0e25f13d349922efd5355f84f2f3a05f947e7171.

PR #159, opened 2019-09-02
Feature/improve visualization

vnmabus:
The plotting functions have been changed to not use global state (using pyplot) whenever possible. Instead, the plotting functions in skfda create and return a new figure, if none is passed.

Pyplot is still used for Sphinx-gallery, as it has to scrap the produced images.

All the examples have been changed in order to use the new API and the object oriented functionalities of Matplotlib.


pablomm:
I've a problem with the new API, when I try this code in a terminal
```python
   import skfda
   a = skfda.datasets.make_sinusoidal_process()
   a.plot().show()
```
The figure it is shown, but the matplotlib window is blocked and can't be closed. I have to close
the terminal.
Dou you have this problem in Linux?

Python 3.6.5 (on darwin)
Matplotlib 2.2.2

vnmabus:
> I've a problem with the new API, when I try this code in a terminal
> 
> ```python
>    import skfda
>    a = skfda.datasets.make_sinusoidal_process()
>    a.plot().show()
> ```
> 
> The figure is shown, but the matplotlib window is blocked and can't be closed. I have to close
> the terminal.
> Dou you have this problem in Linux?
> 
> Python 3.6.5 (on darwin)
> Matplotlib 2.2.2

In Linux it works well for me. Maybe it is for the Mac backend?

I did not like to use `pyplot.figure()` because `pyplot` keeps a global list of figures, and I did not want to modify global state inside a library. However, the documentation in Matplotlib lacks examples of bypassing `pyplot`, and does not provide a simple way of getting the default backend (only provides its name). Thus, my solution is a little bit hackish, although it is well encapsulated and can be changed if we found something better. The backend is added to the figure in a separate step, instead of relying in `pyplot`, so maybe I missed some important step for Mac.

pablomm:
I could solve the problem by changing the backend of matplotlib, in version `2.2.2` the following code shows the window and allows to close it, but the zoom does not work
```python
import matplotlib
matplotlib.use('TkAgg')
import skfda
a = skfda.datasets.make_sinusoidal_process()
fig = a.plot()
fig.show()
```
After upgrading matplotlib to version `3.1.1` the window and buttons works correctly with this backend.
By default it is used the backend `MacOSX` in which the buttons do not work after these changes.

After this I have tried other backends listed in `matplotlib.rcsetup.interactive_bk` and they also work well Qt4Agg and Qt5Agg.

vnmabus:
> I could solve the problem by changing the backend of matplotlib, in version `2.2.2` the following code shows the window and allows to close it, but the zoom does not work
> 
> ```python
> import matplotlib
> matplotlib.use('TkAgg')
> import skfda
> a = skfda.datasets.make_sinusoidal_process()
> fig = a.plot()
> fig.show()
> ```
> 
> After upgrading matplotlib to version `3.1.1` the window and buttons works correctly with this backend.
> By default it is used the backend `MacOSX` in which the buttons do not work after these changes.
> 
> After this I have tried other backends listed in `matplotlib.rcsetup.interactive_bk` and they also work well Qt4Agg and Qt5Agg.

Have you tried defining the environment variable _SKFDA_USE_PYPLOT to 1? If that works then there is some work that pyplot is doing under the hood in that backend which I have not properly replicated.

pablomm:
After  export the variable this code works well.
```python
$ export _SKFDA_USE_PYPLOT=1
$ python
>>> import skfda
>>> import matplotlib.pyplot as plt
>>> skfda.datasets.make_sinusoidal_process().plot()
>>> plt.show()
```
Without the variable, it doesn't show anything.

If the figure is drawn directly, the window does not work correctly (with or without the variable).

```python
$ export _SKFDA_USE_PYPLOT=1
$ python
>>> import skfda
>>> a = skfda.datasets.make_sinusoidal_process()
>>> fig = a.plot()
>>> fig.show()
```

vnmabus:
> After export the variable this code works well.
> 
> ```python
> $ export _SKFDA_USE_PYPLOT=1
> $ python
> >>> import skfda
> >>> import matplotlib.pyplot as plt
> >>> skfda.datasets.make_sinusoidal_process().plot()
> >>> plt.show()
> ```
> 
> Without the variable, it doesn't show anything.

This is expected, as without the environment variable, we are bypassing pyplot.

> If the figure is drawn directly, the window does not work correctly (with or without the variable).
> 
> ```python
> $ export _SKFDA_USE_PYPLOT=1
> $ python
> >>> import skfda
> >>> a = skfda.datasets.make_sinusoidal_process()
> >>> fig = a.plot()
> >>> fig.show()
> ```

If there work the same, I would say that this is a Matplotlib problem. It happens to have several issues in Mac (https://matplotlib.org/3.1.0/faq/osx_framework.html). Thus, unfortunately you would have to set the variable in Mac or pass a previously created figure (with pyplot) as a parameter and then plot with `plt.show()`.

pablomm:
It looks a problem with matplotlib and anaconda in mac. I'll try to fix it.

This changes are a good improvement in the library, but in my opinion the default behaviour of the package should be use ``pyplot``, using the variable to avoid its usage. Although matplotlib offers the object-oriented API, currently, the general use of the library is through ``pyplot``.

I think that users who come from other scikit packages will be used to plotting with pyplot (e.g. all the ``scikit-learn`` examples use it), or other libraries as ``seaborn`` use internally ``pyplot``. 
In addition, other packages not based on matplotlib contain APIs with functionalities similar to pyplot (e.g. ``mayavi.mlab``), and it's a more familiar syntax for users who already know R or MATLAB.

Apart from this, the cases where it is not possible to modify the global variables of pyplot are few, such as the use of plots in GUIs or other particular cases,  in which the option ``_SKFDA_USE_PYPLOT`` is useful, so I think that by default pyplot should continue to be used by default,  and in the future see if it is convenient to change the default behaviour.

vnmabus:
> It looks a problem with matplotlib and anaconda in mac. I'll try to fix it.

Ok, please keep me informed.

> This changes are a good improvement in the library, but in my opinion the default behaviour of the package should be use `pyplot`, using the variable to avoid its usage. Although matplotlib offers the object-oriented API, currently, the general use of the library is through `pyplot`.

Matplotlib recommends the OO API. And, although it recommends `pyplot.figure()` for creating the `Figure` objects, calling the constructor directly is supported. The only hackish bits or my solution are the finding of the default Backend object, for which Matplotlib does not provide an easy way.

> I think that users who come from other scikit packages will be used to plotting with pyplot (e.g. all the `scikit-learn` examples use it), or other libraries as `seaborn` use internally `pyplot`.
> In addition, other packages not based on matplotlib contain APIs with functionalities similar to pyplot (e.g. `mayavi.mlab`), and it's a more familiar syntax for users who already know R or MATLAB.

I do not use Seaborn or Mayavi right now, but they seem to favor an OO API, even if they also provide a stateful one.

> Apart from this, the cases where it is not possible to modify the global variables of pyplot are few, such as the use of plots in GUIs or other particular cases, in which the option `_SKFDA_USE_PYPLOT` is useful, so I think that by default pyplot should continue to be used by default, and in the future see if it is convenient to change the default behaviour.

You can always create a figure with `pyplot.figure()` and pass it to our methods. One particular problem that I had with the stateful API is in the use inside a Jupyter notebook. Jupyter tries to be useful plotting all the plots created from pyplot in each cell, which has the result of some plots appearing twice. Also, it feels wrong that a plot made by a library is exposed in that way.

pablomm:
> I do not use Seaborn or Mayavi right now, but they seem to favor an OO API, even if they also provide a stateful one.

Ok, then I agree with you that we should use the OO API.

> Ok, please keep me informed.

I have read more in depth about Mac problems listed in https://matplotlib.org/3.1.0/faq/osx_framework.html and I think it's because of how the figure is created and not because of the Mac backend.

Effectively, in the documentation they recommend the OO API, but from what I understand of https://matplotlib.org/api/api_overview.html#the-object-oriented-api, when they talk about the pyplot API they mean code like [this](https://matplotlib.org/tutorials/introductory/pyplot.html).

```python
# Simple plot with pyplot API
import matplotlib.pyplot as plt 
plt.plot([1, 2, 3, 4])
plt.ylabel('some numbers')
plt.show()
```
and when they talk about the OO API to code like this
```python
# Same plot with OO API
import matplotlib.pyplot as plt 
fig, ax = plt.subplots()
ax.plot([1, 2, 3, 4])
ax.set_ylabel('some numbers')
plt.show()
```
Both use `plt.figure` and `plt.show` (I checked almost all the examples in https://matplotlib.org/gallery/index.html#examples-index, which they said that are written using the OO API and all use them).

Have you seen an example where they directly create the Figure and show it without using pyplot?

vnmabus:
> Have you seen an example where they directly create the Figure and show it without using pyplot?

Yes, for example https://matplotlib.org/3.1.1/faq/howto_faq.html#matplotlib-in-a-web-application-server.  Another example in https://matplotlib.org/gallery/api/agg_oo_sgskip.html. Also, in the description of the Matplotlib architecture https://aosabook.org/en/matplotlib.html.

pablomm:
> Yes, for example https://matplotlib.org/3.1.1/faq/howto_faq.html#matplotlib-in-a-web-application-server. Another example in https://matplotlib.org/gallery/api/agg_oo_sgskip.html. Also, in the description of the Matplotlib architecture https://aosabook.org/en/matplotlib.html.

But in those examples it is used `savefig` instead of show.

I found this issue (https://github.com/matplotlib/matplotlib/issues/13101) where they talk about the usage of `fig.show`. 

I tried this code in a IPython shell and works well (running it like a script works badly for me)
```python
import matplotlib.pyplot as plt # Running in IPython
plt.figure().show()
```
But this other code shows the window but the close button doesn't work
```python
import skfda # Running in IPython
skfda.datasets.make_sinusoidal_process().plot().show()
```
Besides, when I mix both figures I get an unexpected result.
```python
import matplotlib.pyplot as plt # Running in IPython
import skfda 
plt.figure().show() # Pyplot Figure
skfda.datasets.make_sinusoidal_process().plot().show() # Skfda figure
```
When I press the skfda figure close button the pyplot figure is closed.
Can you try if you get the same results?

vnmabus:
> But in those examples it is used `savefig` instead of show.

That is true.

> I found this issue ([matplotlib/matplotlib#13101](https://github.com/matplotlib/matplotlib/issues/13101)) where they talk about the usage of `fig.show`.
> 
> I tried this code in a IPython shell and works well (running it like a script works badly for me)
> 
> ```python
> import matplotlib.pyplot as plt # Running in IPython
> plt.figure().show()
> ```
> 
> But this other code shows the window but the close button doesn't work
> 
> ```python
> import skfda # Running in IPython
> skfda.datasets.make_sinusoidal_process().plot().show()
> ```

Both work well for me.

> Besides, when I mix both figures I get an unexpected result.
> 
> ```python
> import matplotlib.pyplot as plt # Running in IPython
> import skfda 
> plt.figure().show() # Pyplot Figure
> skfda.datasets.make_sinusoidal_process().plot().show() # Skfda figure
> ```
> 
> When I press the skfda figure close button the pyplot figure is closed.
> Can you try if you get the same results?

Woah, well spotted! I have the same results in Linux. It seems that Matplotlib does not work as I though. I have found [this PR](https://github.com/matplotlib/matplotlib/pull/14024) enlightening. Apparently, I am not the only one that dislikes the global state in pyplot! It seems that there is work in progress to make possible for pyplot to show figures NOT managed by pyplot, but it is not finished yet. I will continue searching for alternatives.

vnmabus:
As the creation of the figure can be changed later in only one place, I have changed it to always use `pyplot.figure()`. In the future, specially if https://github.com/matplotlib/matplotlib/pull/14024 is merged, we should reconsider not using pyplot, but now it seems too troublesome. I did not want to halt progress in the library for this, so I hope that now the PR is ok for you.

examples/plot_landmark_registration.py
```diff
@@ -74,34 +73,36 @@
...
 # Plots warping
-warping.plot()
+fig = warping.plot()
 
 # Plot landmarks
 for i in range(fd.n_samples):
-    plt.scatter([-0.5, 0.5], landmarks[i])
+    fig.axes[0].scatter([-0.5, 0.5], landmarks[i])
+
+fig
```
pablomm:
```suggestion

```

examples/plot_landmark_registration.py
```diff
@@ -74,34 +73,36 @@
...
 # corresponding registered curve as :math:`x^*_i(t) = x_i(h_i(t))`.
 
 fd_registered = fd.compose(warping)
-fd_registered.plot()
+fig = fd_registered.plot()
+
+fig.axes[0].scatter([-0.5, 0.5], [1, 1])
 
-plt.scatter([-0.5, 0.5], [1, 1])
+fig
```
pablomm:
```suggestion

```

examples/plot_pairwise_alignment.py
```diff
@@ -123,98 +123,103 @@
...
+g.plot(fig=fig, color="C1", label="$g_i$")
 
-fd.plot(color="C0", label="$f_i$")
-fig, ax = g.plot(color="C1", label="$g_i$")
+labels = fig.axes[0].get_lines()
+fig.axes[0].legend(handles=[labels[0], labels[-1]])
 
-l = ax[0].get_lines()
-plt.legend(handles=[l[0], l[-1]])
+fig
```
pablomm:
```suggestion

```

examples/plot_pairwise_alignment.py
```diff
@@ -123,98 +123,103 @@
...
 for lam, c in zip(lambdas, color):
-    skfda.preprocessing.registration.elastic_registration_warping(f, g, lam=lam).plot(color=c)
+    skfda.preprocessing.registration.elastic_registration_warping(
+        f, g, lam=lam).plot(fig=fig, color=c)
 
 # Plots identity
-plt.plot(t,t,  color='C0', linestyle="--")
+fig.axes[0].plot(t, t,  color='C0', linestyle="--")
 
+fig
```
pablomm:
```suggestion

```

examples/plot_pairwise_alignment.py
```diff
@@ -123,98 +123,103 @@
...
-f.plot(color='C0', linewidth=2., label='$f$')
-g.plot(color='C1', linewidth=2., label='$g$')
+f.plot(fig=fig, color='C0', linewidth=2., label='$f$')
+g.plot(fig=fig, color='C1', linewidth=2., label='$g$')
 
 # Legend
-plt.legend()
+fig.axes[0].legend()
+
+fig
```
pablomm:
```suggestion

```

vnmabus:
> Yes, I think it's better this way for now, we'll have to keep an eye out for matplotlib updates.
> 
> I don't know if the `fig` sentences added to the examples have some purpose if pyplot is being used (not all the figures of the examples have this sentence and all are being plotted), if they have, ignore the change suggestions.

They were added in order to reproduce the examples using a notebook if pyplot is not used. With pyplot they are probably printed twice, so I am not sure if we should keep them or not.

pablomm:
> They were added in order to reproduce the examples using a notebook if pyplot is not used. With pyplot they are probably printed twice, so I am not sure if we should keep them or not.

In that case I think we should remove them while pyplot is in use.

vnmabus:
I have applied your suggestions.

PR #161, opened 2019-09-07
Unify scalar and functional regressors

### LINKED ISSUE #158: Unify scalar and functional regressors
I was mentioned in a meeting that it makes no sense to differentiate between scalar and functional regressors.  First, because there are not scalar, but multivariate, and second, because there is no confusion: the output of `predict` depends on the input to `fit`.

I forgot to mention this on the review of #112, but it should not be difficult to change. Can you do it @pablomm ?

pablomm:
* Unify ``NeighborsScalarRegressor`` and ``NeighborsFunctionalRegressor`` in the same class.
* Fix links to examples in neighbors documentation.
* Rename `sklearn_metric` parameter in the neighbors estimator to a proper name ``multivariate_metric``.

Closes #158.


pablomm:
> I have a question. Why is `local_regressor` public and assigned on fit? Should it not be called `local_regressor_` in that case?

Mmm, you are right, it is not a good name, but maybe should be called ``_local_regressor``, since it is for private use and not uses any information of the train data.

vnmabus:
> 
> 
> > I have a question. Why is `local_regressor` public and assigned on fit? Should it not be called `local_regressor_` in that case?
> 
> Mmm, you are right, it is not a good name, but maybe should be called `_local_regressor`, since it is for private use and not uses any information of the train data.

Yes, if it is private it should be `_local_regressor`.

examples/plot_neighbors_functional_regression.py
```diff
@@ -11,20 +11,21 @@
...
 
 ##############################################################################
 #
 # In this example we are going to show the usage of the nearest neighbors
 # regressors with functional response. There is available a K-nn version,
-# :class:`~skfda.ml.regression.KNeighborsFunctionalRegressor`, and other one
-# based in the radius,
-# :class:`~skfda.ml.regression.RadiusNeighborsFunctionalRegressor`.
+# :class:`KNeighborsRegressor
+# <skfda.ml.regression.KNeighborsRegressor>`, and other one based in
```
vnmabus: Use the abbreviated notation (`~path.to.class`) unless you do not want to use the name of the class in the link.

skfda/_neighbors/base.py
```diff
@@ -508,14 +483,15 @@ def fit(self, X, y):
...
-            if not self.sklearn_metric:
+            if not self.multivariate_metric:
 
                 if self.metric == 'l2':
                     metric = lp_distance
                 else:
                     metric = self.metric
 
-                sk_metric = _to_sklearn_metric(metric, self._sample_points)
+                sk_metric = _to_multivariate_metric(metric,
```
vnmabus: Maybe `sk_metric` should change its name too. With this change it could be confusing now.

skfda/_neighbors/classification.py
```diff
@@ -351,9 +353,10 @@ class and return a :class:`FData` object with only one sample
     --------
     KNeighborsClassifier
     RadiusNeighborsClassifier
-    KNeighborsScalarRegressor
-    RadiusNeighborsScalarRegressor
+    KNeighborsRegressor
+    RadiusNeighborsegressor
```
vnmabus:
```suggestion
    RadiusNeighborsRegressor
```

pablomm: Done in 07083ef.

examples/plot_neighbors_functional_regression.py
pablomm: Done in 07083ef.

skfda/_neighbors/base.py
pablomm: Some like ``sklearn_multivariate_metric``?  (``multivariate_metric`` it is used as an argument of the estimator)

vnmabus: Maybe that is too long. But at least `sklearn_metric`, since the sk prefix is common for many scikits.

pablomm: Applied in c1fbcc3.

PR #171, opened 2019-10-14
Transformers to multivariate variables

vnmabus:
Adds two transformers (`CoefficientsTransformer` and `EvaluationTransformer`) that convert functional data into multivariate data that can be further processed with scikit-learn.


vnmabus:
> Ok, seems right.
> I don't know if intended, but the transformers are not added to the documentation.

I forgot to add it. I will add it in a future PR (maybe with an example showing integration with scikit-learn estimators in the same pipeline).

PR #173, opened 2019-10-31
Changes to plotting groups

### LINKED ISSUE #165: Enhancement  in sample_labels
The current ``sample_labels`` argument of ``plot`` to group samples by colors is very restrictive, only accepts labels of the form *0, 1, 2, ... n_classes -1*  without skip any number.

Can be fixed using [LabelEncoder](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.LabelEncoder.html) internally.

**To Reproduce**
Code to reproduce the behavior:

```python
from skfda.datasets import make_sinusoidal_process
fd = make_sinusoidal_process(n_samples=5)

fd.plot(sample_labels=[0, 0, 2, 2, 2])  # Not valid
fd.plot(sample_labels=[-1, -1, 0, 1, 1]) # Not valid
```
**Result**
> **ValueError**: sample_labels must contain at least an occurence of numbers between 0 and number of distint sample labels.

vnmabus:
* Rename sample_labels to group.
* Rename label_colors to group_colors.
* Rename label_names to group_names.
* `group` is now not required to be an array of integers, closes #165.
* `group_colors` and `group_names` are now objects that index the
color/name per group (so if `group` is an array of integers, they can
still be arrays, but they can be dict-like objects for other types).
* A new parameter `legend` allows to plot a legend even if no
`group_names` is passed (it will use the objects passed to `group`).
* The *explore* example is changed to make use of some of these features
in order to test them.

### DIFF
--- a/examples/plot_boxplot.py
+++ b/examples/plot_boxplot.py
@@ -38,9 +38,9 @@ label_names = dataset["target_names"]
 nlabels = len(label_names)
 label_colors = colormap(np.arange(nlabels) / (nlabels - 1))
 
-fd_temperatures.plot(sample_labels=dataset["target"],
-                     label_colors=label_colors,
-                     label_names=label_names)
+fd_temperatures.plot(group=dataset["target"],
+                     group_colors=label_colors,
+                     group_names=label_names)
 
 
 ##############################################################################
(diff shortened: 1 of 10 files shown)


vnmabus:
> As in the code the `get_cmap` function is used to get the default colormap, I don't know if this is the expected behaviour or a bug.

Amanda choosed that behaviour on her own. I think that we do not specify *what* colors we use, only that they are different. But probably it is better to use the default color cycle in this case. I will update this PR.

PR #188, opened 2020-03-19
Feature/fpca

hzzhyj:
This branch implements the main features of FPCA. It includes:

1. FPCA for functions in a discretized form.
2. FPCA for functions in a basis form.
3. Example of execution
4. Documentation for the mentioned functionality

I have separated the functionality about finding the best regularization parameter in the following branch, in order to make smaller branches.


hzzhyj:
The failing builds come from the doctests. This is because the principal components obtained (the vectors) can be the vector or the negated vector. Maybe we can remove this comparison as it is not relevant and we only want to show usage?

docs/modules/exploratory.rst
```diff
@@ -10,4 +10,5 @@ and visualize functional data.
    
    exploratory/visualization
    exploratory/depth
-   exploratory/outliers
\ No newline at end of file
+   exploratory/outliers
+   exploratory/fpca
```
vnmabus: FPCA is not a main category. It should be in preprocessing/dim_reduction/projection in my opinion

docs/modules/exploratory/fpca.rst
```diff
@@ -0,0 +1,30 @@
...
+This module provides tools to analyse functional data using FPCA. FPCA is
+a common tool used to reduce dimensionality while preserving the maximum
+quantity of variance in the data. FPCA be applied to a functional data object
+in either a basis representation or a discretized representation. The output
+of FPCA are orthogonal functions (usually a much smaller sample than the input
+data sample) that represent the most important modes of variation in the
+original data sample.
+
+For a detailed example please view `FPCA example
+<../../auto_examples/plot_fpca.html>`_, where the process is applied to several
```
vnmabus: I recall there was a better way to link to examples

skfda/representation/basis.py
```diff
@@ -1482,6 +1482,17 @@ def penalty(self, derivative_degree=None, coefficients=None):
             # implement using inner product
             return self._numerical_penalty(coefficients)
 
+    def gram_matrix(self):
```
vnmabus: This is only true if the period and the domain range are the same. Otherwise the basis is not orthonormal

tests/test_fpca.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def fetch_weather_temp_only():
+    weather_dataset = fetch_weather()
+    fd_data = weather_dataset['data']
+    fd_data.data_matrix = fd_data.data_matrix[:, :, :1]
+    fd_data.axes_labels = fd_data.axes_labels[:-1]
+    return fd_data
+
+class MyTestCase(unittest.TestCase):
```
vnmabus: Improve naming

tests/test_fpca.py
```diff
@@ -0,0 +1,90 @@
...
+        results = np.array(results)
+
+        # compare results obtained using this library. There are slight
+        # variations due to the fact that we are in two different packages
+        for i in range(n_components):
+            if np.sign(fpca.components.coefficients[i][0]) != np.sign(results[i][0]):
+                results[i, :] *= -1
+            for j in range(n_basis):
+                self.assertAlmostEqual(fpca.components.coefficients[i][j],
+                                       results[i][j], delta=0.0000001)
```
vnmabus: Is it necessary to specify the delta?

docs/modules/exploratory.rst
hzzhyj: As it is a technique used for dimension reduction I see what you mean. It is basically a special type of projection. I agree with you and I'll soon submit another commit with file locations changed.

hzzhyj: And also, there are some FPCA code already in the same directory, but I believe that the FPCA done there is incorrect as it does not take into account the weight matrix. Should I remove that file?

skfda/representation/basis.py
hzzhyj: I see. So check if period and domain_range are the same. If so return identity and otherwise return super.gram_matrix()?

tests/test_fpca.py
hzzhyj: Yes, as due to precision reasons that the delta we obtain is this one. If we do not specify delta, the comparison goes beyond that precision and the test case fails

skfda/representation/basis.py
vnmabus: That sounds good. Remember that the period must be equal to the difference of the limits in the domain range.

docs/modules/exploratory.rst
vnmabus: Yes, remove it.

docs/modules/exploratory/fpca.rst
hzzhyj: Yes, I am changing to that way right now.

hzzhyj: I used this one because it is the first one I found, but after viewing other examples I found the one you mentioned that does not depend on the directory we are.

examples/plot_fpca.py
```diff
@@ -0,0 +1,113 @@
...
+                                  mean_fd.coefficients[0, :] -
+                                  20 * fpca.components.coefficients[1, :]])
+mean_fd.plot()
+
+##############################################################################
+# We can also specify another basis for the principal components as argument
+# when creating the FPCABasis object. For example, if we use the Fourier basis
+# for the obtained principal components we can see that the components are
+# periodic. This example is only to illustrate the effect. In this dataset, as
+# the functions are not periodic it does not make sense to use the Fourier basis
```
vnmabus: Then, why not using monomials as the other example?

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,425 @@
...
+            n_components (int): number of principal components to obtain from
+                functional principal component analysis
+            centering (bool): if True then calculate the mean of the functional
+                data object and center the data first. Defaults to True
+        """
+        self.n_components = n_components
+        self.centering = centering
+        self.components = None
+        self.component_values = None
+        self.pca = PCA(n_components=self.n_components)
```
vnmabus: The scikit-learn guidelines recommend that the __init__ method should be "dumb", and the checks and transformations of the arguments should be done in the `fit` method. Also attributes set in `fit` should end in a trailing underscore.

docs/modules/preprocessing/dim_reduction/fpca.rst
```diff
@@ -0,0 +1,30 @@
...
+quantity of variance in the data. FPCA be applied to a functional data object
+in either a basis representation or a discretized representation. The output
+of FPCA are orthogonal functions (usually a much smaller sample than the input
+data sample) that represent the most important modes of variation in the
+original data sample.
+
+For a detailed example please view :ref:`sphx_glr_auto_examples_plot_fpca.py`,
+where the process is applied to several datasets in both discretized and basis
+forms.
+
```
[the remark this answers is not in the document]
hzzhyj: The problem with that approach is that too much functionality will be included in the same class, and the fit method will become too large. It may be very bad for future maintenance. 
Second, there are attributes that are exclusively for FPCABasis and attributes that are exclusively for FPCADiscretized. 
Having too much functionality in the same class may also confuse the users about what attributes to use when they want to apply it to a FDadaGrid compared to when they want to apply it for a FDataBasis.
Lastly, regularization is only available for FDataBasis. Which is another attribute that distinguishes both estimators.

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: Fixed. 
However, when passing the check function for estimators, it treats X as numpy ndarray and does not accept that I am using attributes from FDataBasis. Is there a fix to that?

vnmabus: Do not check using the functions from scikit-learn. Maybe we should have our own checking functions for estimators, but for now do not perform any check.

examples/plot_fpca.py
```diff
@@ -0,0 +1,113 @@
...
+# We initialize the FPCABasis object and run the fit function to obtain the
+# first 2 principal components. By default the principal components are
+# expressed in the same basis as the data. We can see that the obtained result
+# is similar to the discretized case.
+fpca = FPCABasis(n_components=2)
+fpca.fit(basis_fd)
+fpca.components.plot()
+
+##############################################################################
+# To better illustrate the effects of the obtained two principal components,
```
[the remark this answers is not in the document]
hzzhyj: I think so because it is a very common way to illustrate the effects of the principal components. 
Maybe we can include a method that returns a new FDataBasis or FDataGrid with the mean and variations over the mean?

skfda/preprocessing/dim_reduction/__init__.py
```diff
@@ -0,0 +1 @@
+from . import projection
```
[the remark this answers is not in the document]
hzzhyj: done

docs/modules/preprocessing.rst
```diff
@@ -28,4 +29,14 @@ Sometimes, the functional data may be misaligned, or the phase variation
 should be ignored in the analysis. To align the data and eliminate the phase
 variation, we need to use *registration* methods. 
 :doc:`Here <preprocessing/registration>` you can learn more about the
-registration methods available in the library.
\ No newline at end of file
+registration methods available in the library.
+
+Dimension Reduction
```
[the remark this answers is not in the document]
hzzhyj: Agreed

examples/plot_fpca.py
hzzhyj: This is to illustrate that we can change the basis but we have to be careful about it.
It is a good idea to further illustrate the change of basis with a monomial basis and show that the effect is much better.

docs/modules/preprocessing/dim_reduction/fpca.rst
```diff
@@ -0,0 +1,30 @@
+Functional Principal Component Analysis (FPCA)
+==============================================
+
+This module provides tools to analyse functional data using FPCA. FPCA is
+a common tool used to reduce dimensionality while preserving the maximum
+quantity of variance in the data. FPCA be applied to a functional data object
+in either a basis representation or a discretized representation. The output
+of FPCA are orthogonal functions (usually a much smaller sample than the input
```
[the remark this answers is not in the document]
hzzhyj: I might use some sleep too...

docs/modules/preprocessing/dim_reduction/fpca.rst
```diff
@@ -0,0 +1,32 @@
...
+
+   skfda.preprocessing.dim_reduction.projection.FPCABasis
+
+FPCA for functional data in a discretized representation
+----------------------------------------------------------------
+
+.. autosummary::
+   :toctree: autosummary
+
+   skfda.preprocessing.dim_reduction.projection.FPCADiscretized
```
vnmabus: Why not FPCAGrid?

docs/modules/preprocessing/dim_reduction/fpca.rst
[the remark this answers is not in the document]
vnmabus: Im still not convinced, but we can always change it later, if necessary.

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,433 @@
...
+    in basis form.
+
+    Attributes:
+        n_components (int): number of principal components to obtain from
+            functional principal component analysis. Defaults to 3.
+        centering (bool): if True then calculate the mean of the functional data
+            object and center the data first. Defaults to True. If True the
+            passed FDataBasis object is modified.
+        components_ (FDataBasis): this contains the principal components either
+            in a basis form.
```
vnmabus: The sentence is incomplete.

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,433 @@
...
 
-    Returns:
-        tuple: (scores, principal directions, eigenvalues)
+
+class FPCABasis(FPCA):
+    """Funcional principal component analysis for functional data represented
+    in basis form.
+
+    Attributes:
+        n_components (int): number of principal components to obtain from
```
vnmabus: The attributes set in `__init__` should be parameters, not attributes.

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,433 @@
...
+            functional principal component analysis. Defaults to 3.
+        centering (bool): if True then calculate the mean of the functional data
+            object and center the data first. Defaults to True. If True the
+            passed FDataBasis object is modified.
+        components_ (FDataBasis): this contains the principal components either
+            in a basis form.
+        components_basis (Basis): the basis in which we want the principal
+            components. We can use a different basis than the basis contained in
+            the passed FDataBasis object.
+        component_values_ (array_like): this contains the values (eigenvalues)
```
vnmabus: Why not `explained_variance_`, as in scikit-learn?

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,433 @@
...
+            associated with the principal components.
+        pca_ (sklearn.decomposition.PCA): object for PCA.
```
vnmabus: I think this should be private, as it is an implementation detail.

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,433 @@
...
+        References:
+            .. [RS05-8-4-2] Ramsay, J., Silverman, B. W. (2005). Basis function
+                expansion of the functions. In *Functional Data Analysis*
+                (pp. 161-164). Springer.
+
+        """
+
+        # the maximum number of components is established by the target basis
+        # if the target basis is available.
+        n_basis = self.components_basis.n_basis if self.components_basis \
```
vnmabus: Prefer parens rather than `\`.

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,433 @@
...
+            g_matrix = self.components_basis.gram_matrix()
+            j_matrix = g_matrix
+
+        # make g matrix symmetric, referring to Ramsay's implementation
+        g_matrix = (g_matrix + np.transpose(g_matrix)) / 2
+
+        # Apply regularization / penalty if applicable
+        if self.regularization_parameter > 0:
+            # obtain regularization matrix
+            regularization_matrix = self.components_basis.penalty(
```
vnmabus: This call is going to change with #189 .

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,433 @@
...
+                regularization). When this value is large, the resulting
+                principal components tends to be 0.
+
+        """
+        super().__init__(n_components, centering)
+        # basis that we want to use for the principal components
+        self.components_basis = components_basis
+        # lambda in the regularization / penalization process
+        self.regularization_parameter = regularization_parameter
+        self.regularization_derivative_degree = regularization_derivative_degree
```
vnmabus: After #189, we will favour receiving a single parameter which will be used to initialize a LinearDifferentialOperator (if it is not one already).

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,433 @@
...
+                * regularization_matrix
+
+        # obtain triangulation using cholesky
+        l_matrix = np.linalg.cholesky(g_matrix)
+
+        # we need L^{-1} for a multiplication, there are two possible ways:
+        # using solve to get the multiplication result directly or just invert
+        # the matrix. We choose solve because it is faster and more stable.
+        # The following matrix is needed: L^{-1}*J^T
+        l_inv_j_t = np.linalg.solve(l_matrix, np.transpose(j_matrix))
```
vnmabus: Is it not `scipy.linalg.cho_solve` faster?

docs/modules/preprocessing/dim_reduction/fpca.rst
hzzhyj: Agreed, this name is much more straightforward and user friendly.

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: I noticed that and it is on my todo list, just waiting for your pull request to merge and then I will change it to a single parameter.

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: I forgot to change this to solve_triangular, which is the fastest of the methods I have compared.

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: Thanks for noticing that.

examples/plot_fpca.py
hzzhyj: Ill use monomials as another example and illustrate it, didn't get your point the other day.

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: Extra computation is needed to obtain the principal components. Maybe we can expose pca_ so the user can choose what he wants to obtain from this module? As pca from scikit learn provides more attributes that the user may be interested in.

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: Solved

examples/plot_fpca.py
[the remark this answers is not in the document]
vnmabus: Put this into a new issue.

skfda/exploratory/fpca/test.ipynb
```diff
@@ -0,0 +1,3059 @@
+{
```
vnmabus, line 1: Remove this file from the commit history.

tests/test_fpca.py
```diff
@@ -0,0 +1,82 @@
...
+        with self.assertRaises(AttributeError):
+            fpca.fit(fd)
+
+        # check that n_components must be smaller than the number of attributes
+        # in the FDataGrid object
+        fd = FDataGrid([[0.9], [0.7], [0.5]], sample_points=[0])
+        with self.assertRaises(AttributeError):
+            fpca.fit(fd)
+
+    def test_basis_fpca_fit_result(self):
```
vnmabus, line 47: What about a similar test for the discretized form?

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,445 @@
...
+        components_ (FDataBasis): this contains the principal components in a
+            basis representation.
+        component_values_ (array_like): this contains the values (eigenvalues)
+            associated with the principal components.
+        pca_ (sklearn.decomposition.PCA): object for PCA.
+            In both cases (discretized FPCA and basis FPCA) the problem can be
+            reduced to a regular PCA problem and use the framework provided by
+            sklearn to continue.
+
+    Parameters:
```
vnmabus, line 103: Reverse the order: first parameters, then attributes.

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,445 @@
...
+class FPCABasis(FPCA):
+    """Functional principal component analysis for functional data represented
+    in basis form.
+
+    Attributes:
+        components_ (FDataBasis): this contains the principal components in a
+            basis representation.
+        component_values_ (array_like): this contains the values (eigenvalues)
+            associated with the principal components.
+        pca_ (sklearn.decomposition.PCA): object for PCA.
```
vnmabus, line 98: This should not be public.

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,445 @@
...
+
+
+class FPCABasis(FPCA):
+    """Functional principal component analysis for functional data represented
+    in basis form.
+
+    Attributes:
+        components_ (FDataBasis): this contains the principal components in a
+            basis representation.
+        component_values_ (array_like): this contains the values (eigenvalues)
```
vnmabus, line 96: Why not `explained_variance_`?

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,445 @@
...
+                want the principal components. Defaults to None. If so, the
+                basis contained in the passed FDataBasis object for the fit
+                function will be used.
+            centering (bool): if True then calculate the mean of the functional
+                data object and center the data first. Defaults to True
+            regularization_parameter (float): this parameter sets the degree of
+                regularization that is desired. Defaults to 0 (no
+                regularization). When this value is large, the resulting
+                principal components tends to be constant.
+            regularization_lfd (LinearDifferentialOperator, list or int): Linear
```
vnmabus, line 154: We should try to use the same nomenclature whenever possible in `BasisSmoother`, `MultivariateLinearRegression` and here.

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,445 @@
...
+
+    """
+
+    def __init__(self,
+                 n_components=3,
+                 components_basis=None,
+                 centering=True,
+                 regularization_parameter=0,
+                 regularization_lfd=2):
+        """FPCABasis constructor
```
vnmabus, line 139: I think that it is better not to repeat the parameters in the docstring of the `__init__` method, as scikit-learn does.

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,445 @@
...
+        # in FDataBasis
+        if self.centering:
+            meanfd = X.mean()
+            # consider moving these lines to FDataBasis as a centering function
+            # subtract from each row the mean coefficient matrix
+            fd_data -= np.squeeze(meanfd.data_matrix)
+
+        # establish weights for each point of discretization
+        if not self.weights:
+            # sample_points is a list with one array in the 1D case
```
vnmabus, line 406: Is it trapezoidal by default?

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,445 @@
...
-
-    return scores, principal_directions, eigenvalues
+
+    def __init__(self, n_components=3, weights=None, centering=True):
+        """FPCABasis constructor
+
+        Args:
+            n_components (int): number of principal components to obtain from
+                functional principal component analysis
+            weights (numpy.array): the weights vector used for discrete
```
vnmabus, line 340: I think this could also be a callable.

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,445 @@
...
+            fd_data -= np.squeeze(meanfd.data_matrix)
+
+        # establish weights for each point of discretization
+        if not self.weights:
+            # sample_points is a list with one array in the 1D case
+            # in trapezoidal rule, suppose \deltax_k = x_k - x_{k-1}, the weight
+            # vector is as follows: [\deltax_1/2, \deltax_1/2 + \deltax_2/2,
+            # \deltax_2/2 + \deltax_3/2, ... , \deltax_n/2]
+            differences = np.diff(X.sample_points[0])
+            self.weights = [sum(differences[i:i + 2]) / 2 for i in
```
vnmabus, line 411: Vectorize this.

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,445 @@
...
+                only present because of fit function convention
+
+        Returns:
+            (array_like): the scores of the data with reference to the
+            principal components
+        """
+
+        # in this case its the coefficient matrix multiplied by the principal
+        # components as column vectors
+        return X.copy(data_matrix=np.squeeze(X.data_matrix) @ np.transpose(
```
vnmabus, line 444: Why do you need `squeeze`? I prefer to know the dimensions of the data.

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: Yes, it is trapezoidal by default

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: If we pass the data matrix directly, the dimension is NxMx1, as a 3D array. Therefore we have to remove the inner-most dimension. Maybe you have a better method of doing this?

vnmabus: If it is necessary to drop dimensions I prefer `reshape`, in order to be sure of which dimensions are kept.

tests/test_fpca.py
hzzhyj: Done, it took me some time to study fda.usc, but better late than never right

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: Done

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: Yeah that is better, done

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: Yes, it should not be public, removed from attributes

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: I think that explained_variance_ratio is better, as it gives the percentage which is more interesting

vnmabus: Why not both?

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: Done

examples/plot_fpca.py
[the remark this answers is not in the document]
hzzhyj: Done

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: Not sure why we should vectorize this as the sample points usually are small?

vnmabus: Because it is trivial and more simple:

```
differences = np.concatenate(((0,), differences , (0,)))
self.weights = (differences [:-1] + differences [1:]) / 2
```

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: Yeah it costs nothing and both may be interesting

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: Done

skfda/preprocessing/dim_reduction/projection/_fpca.py
hzzhyj: Calling FPCAGrid on a FDataGrid object? with default parameters or the user may also choose the parameters?

skfda/preprocessing/dim_reduction/projection/_fpca.py
vnmabus: This still has a different name

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,532 @@
...
+            object and center the data first. Defaults to True. If True the
+            passed FDataBasis object is modified.
+        weights (numpy.array Union callable): the weights vector used for
+            discrete integration. If none then the trapezoidal rule is used for
+            computing the weights. If a callable object is passed, then the
+            weight vector will be obtained by evaluating the object at the
+            sample points of the passed FDataGrid object in the fit method.
+        regularization_parameter (float): this parameter determines the amount
+            of smoothing applied. Defaults to 0
+        penalty (Union[int, Iterable[float]): the coefficients that will be
```
vnmabus, line 369:
```suggestion
        penalty (Union[int, Iterable[float]]): the coefficients that will be
```
although if you want to be so specific with types you can use annotations directly.

skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -1,33 +1,532 @@
...
+    """Funcional principal component analysis for functional data represented
+    in discretized form.
+
+    Parameters:
+        n_components (int): number of principal components to obtain from
+            functional principal component analysis. Defaults to 3.
+        centering (bool): if True then calculate the mean of the functional data
+            object and center the data first. Defaults to True. If True the
+            passed FDataBasis object is modified.
+        weights (numpy.array Union callable): the weights vector used for
```
vnmabus, line 362:
```suggestion
        weights (numpy.array or callable): the weights vector used for
```
(or use the Union notation properly).

skfda/preprocessing/dim_reduction/projection/_fpca.py
vnmabus: What is the difference between `component_values_ ` and `explained_variance_ `?

hzzhyj: The difference is that explained_variance_ = (singular_values ** 2) / (n_samples - 1)

where singular values**2 = component_values

I have removed component_values as it is redundant in this case.

PR #189, opened 2020-03-21
Simplify basis

### LINKED ISSUE #185: B spline penalty matrix is wrong
**Describe the bug**
0 order B spline penalty matrix should be non-negative since its function values are all negative. However, I get penalty matrix with negative values.

**To Reproduce**
Code to reproduce the behavior:

```python
from skfda.representation.basis import BSpline
bss = BSpline(n_basis=16, order=8)
bss.penalty(0)
```

**Expected behavior**
A non-negative and semi-positive definite matrix

**Version information**
 - OS: Windows
 - Python version: 3.6
 - scikit-fda version: 0.3
 - Version of other packages involved [None ]

**Additional context**

vnmabus:
Several simplifications and improvements for basis representation module:
- Split FDatabasis to another module for easier manipulation.
- Evaluation methods for existin basis have been simplified when possible.
- The `penalty` method now accepts a linear differential operator.
- The `_penalty` method now is the method to override for more accurate computation of the penalty matrix. If it is not overriden, or it returns `NotImplemented`, a numerical computation is done instead.
- The numerical computation of the penalty matrix is now vectorized and simplified.
- The _penalty matrices for other basis have been reworked and generalized (all but the BSpline basis) to constant weight linear differential operators.
- A bug has been fixed in the BasisSmoother which prevented complex linear differential operators to be used.
- The `n_degenerated` method is dropped for now. It was only used if the method was QR, and was not general enough.
- As a result of computing the penalty using the original (translated) coordinates for the piecewise polynomials in the BSpline penalty, instead of converting them to the standard power basis, the accuracy has been increased, and it closes #185 .
- More tests have been added.


skfda/misc/_lfd.py
```diff
@@ -165,24 +176,42 @@ def __init__(self, order=None, *, weights=None, domain_range=None):
...
     def __eq__(self, other):
         """Equality of Lfd objects"""
-        return (self.order == other.nderic and
-                all(self.weights[i] == other.bwtlist[i]
-                    for i in range(self.order)))
+        return (self.weights == other.weights)
+
+    def constant_weights(self):
+        """
+        Return the weights of the weights if they are constant basis.
```
hzzhyj: What does this phrase mean?

vnmabus: Probably means that I need more sleep. Thanks for noticing it.

vnmabus: Fixed in 4fb0d589437d061c4ad723cfd503b9e24b663a0a.

skfda/representation/basis.py
```diff
@@ -1340,69 +1236,102 @@ def _compute_matrix(self, eval_points, derivative=0):
...
+
+    def _penalty(self, lfd):
+
+        weights = lfd.constant_weights()
+        if weights is None:
+            return NotImplemented
+
+        # If the period and domain range are not the same, the basis functions
+        # are not orthogonal
+        if self.period != (self.domain_range[0][1] - self.domain_range[0][0]):
```
hzzhyj: In this case, can´t we use inner_product to compute the penalty matrix?

vnmabus: When the function returns NotImplemented, it will do a numerical approach. For orthonormal Fourier basis, computing the penalty matrix when the linear differential operator has constant weights is easily done analytically, but I am not sure of how it should be done in other cases.

The numerical approach cannot use the inner product code (unless we generalize this code to accept arbitrary callables) as the inner product inside is not between things expressed in basis nor in grid form.

PR #200, opened 2020-04-21
extract depth based median from boxplot

hzzhyj:
As the computation of the depth measure is still necessary in boxplot, I haven't removed the computation of the depth median from boxplot. This is just a simple extraction of the depth-based median.

### DIFF
--- a/skfda/exploratory/stats/_stats.py
+++ b/skfda/exploratory/stats/_stats.py
@@ -1,6 +1,6 @@
 """Functional data descriptive statistics.
 """
-
+from ..depth import modified_band_depth
 
 def mean(fdata, weights=None):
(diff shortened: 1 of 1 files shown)


skfda/exploratory/stats/_stats.py
```diff
@@ -67,3 +67,27 @@ def cov(fdatagrid):
 
     """
     return fdatagrid.cov()
+
+
+def depth_based_median(fdatagrid, depth_method=modified_band_depth):
+    """Compute the median based on a depth measure.
+
+    The depth based median is basically the deepest curve given a certain
```
vnmabus, line 75:
```suggestion
    The depth based median is the deepest curve given a certain
```

PR #204, opened 2020-04-27
trimmed means

hzzhyj:
The trimmed means is the mean of the n-n\alfa deepest curves, given a certain depth measure. This has been implemented in exploratory.stats


skfda/exploratory/stats/_stats.py
```diff
@@ -91,3 +91,38 @@ def depth_based_median(fdatagrid, depth_method=modified_band_depth):
 
     # The median is the deepest curve
     return fdatagrid[indices_descending_depth[0]]
+
+
+def trimmed_means(fdatagrid,
```
vnmabus, line 96: Maybe should we use same name of function and parameters as numpy (https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.trim_mean.html)?

hzzhyj: done

skfda/exploratory/stats/_stats.py
```diff
@@ -91,3 +91,38 @@ def depth_based_median(fdatagrid, depth_method=modified_band_depth):
...
+    """Compute the trimmed means based on a depth measure.
+
+    The trimmed means consists in computing the mean function without a
+    percentage of least deep curves. That is, we first remove the least deep
+    curves and then we compute the mean as usual.
+
+    Args:
+        fdatagrid (FDataGrid): Object containing different samples of a
+            functional variable.
+        proportiontocut (float): indicates the percentage of functions to
```
vnmabus, line 108: There is a problem I have just noticed. In numpy and scipy, `proportiontocut` refers to both sides, while here we only have one side.

skfda/exploratory/stats/_stats.py
```diff
@@ -91,3 +91,38 @@ def depth_based_median(fdatagrid, depth_method=modified_band_depth):
...
+            dataset.
+        depth_method (:ref:`depth measure <depth-measures>`, optional):
+            Method used to order the data. Defaults to :func:`modified
+            band depth <fda.depth_measures.modified_band_depth>`.
+
+    Returns:
+        FDataGrid: object containing the computed trimmed mean.
+
+    """
+    n_samples_to_keep = int((fdatagrid.n_samples -
```
vnmabus, line 119: You can use `scipy.stats.trim1` to simplify the code even more and make its behaviour as similar as possible with scipy.

skfda/exploratory/stats/_stats.py
hzzhyj: The thing is that with functional data it is not possible to cut on the other direction right? we are ranking the data according to their depth measure.

skfda/exploratory/stats/_stats.py
hzzhyj: I tried this but FDataGrid is not compatible with the internal operations of trim1. Maybe just use what is implemented as it is trivial?

vnmabus: Ok, but make sure that the rounding is the same.

skfda/exploratory/stats/_stats.py
vnmabus: Right. The problem is the possible confusion for a person used to the scipy `trim_mean`. But the only thing that I can think of is to make a note in the documentation explaining it.

skfda/exploratory/stats/_stats.py
hzzhyj: Done, i am using the exact same operation now to find the index where we  cut.

skfda/exploratory/stats/_stats.py
hzzhyj: Added the note to the documentation. I think that after all we are treating different types of data and the people using this library should be aware. But more documentation never hurts

vnmabus: No, I think the note misses the point. The important thing to note here is that the scipy function removes twice the `proportiontocut` while we remove it once.

hzzhyj: That is true, it does not make sense to remove twice the data in our case as it is unidirectional.

PR #209, opened 2020-05-16
perturbations over mean

hzzhyj:
This new method computes the perturbations over the mean and return them as FDataBasis or FDataGrid depending on the case.


skfda/preprocessing/dim_reduction/projection/_fpca.py
```diff
@@ -397,3 +397,39 @@ def fit_transform(self, X, y=None, **fit_params):
         """
         self.fit(X, y)
         return self.transform(X, y)
+
+    def get_component_perturbations(self, X, index=0, multiple=30):
```
vnmabus, line 401: Why is this a method of FPCA?

vnmabus, line 401: Why `30` as the default value?

hzzhyj: This is just an arbitrary value, I was thinking about making this attribute not optional. This parameter depends on each dataset and cannot be predetermined.

hzzhyj: In this case just to access the components. It can also be a static method outside of FPCA. Making it part of FPCA just makes it easier for the user, in my opinion.

vnmabus: I think that is better to make a function in the visualization module, which accepts as parameters the mean and an iterable of components, making a subplot for each component.

hzzhyj: skfda.exploratory.visualization fpca.py?
I agree with you about making a method to plot the diagram for each component, I'm not sure if this is the location where I should put it.

vnmabus: That location looks good to me.

hzzhyj: Done

skfda/exploratory/visualization/fpca.py
```diff
@@ -0,0 +1,67 @@
+from matplotlib import pyplot as plt
+from skfda.representation import FDataGrid, FDataBasis, FData
+
+
+def plot_fpca_perturbation_graphs(mean, components, multiple,
```
vnmabus, line 5: Try to keep the same API as other visualization functions. Use `_get_figure_and_axes` to parse arguments.

skfda/exploratory/visualization/fpca.py
```diff
@@ -0,0 +1,67 @@
...
+    Returns:
+        (FDataGrid or FDataBasis): this contains the mean function followed
+        by the positive perturbation and the negative perturbation.
+    """
+    if fig is None:
+        fig = plt.figure(figsize=(6, 4 * len(components)))
+    axes = fig.subplots(nrows=len(components))
+
+    for i in range(len(axes)):
+        aux = _get_component_perturbations(mean, components, i, multiple)
```
vnmabus, line 35: If `len(mean) > 1` I would compute the mean. It is easier for the user.

skfda/preprocessing/dim_reduction/projection/_fpca.py
vnmabus, line 401: Remove it.

skfda/exploratory/visualization/fpca.py
hzzhyj: Agreed and done

skfda/exploratory/visualization/fpca.py
hzzhyj: Done

PR #211, opened 2020-05-23
fpca transform method quick fix

hzzhyj:
Transform method bug fixed


hzzhyj:
> If there was a bug and the tests did not catch it, include a new test that does.

Done

PR #220, opened 2020-06-07
Change default interpolation

vnmabus:
Adds linear interpolation. For derivatives, this interpolator interpolates (again, linearly) the finite differences.

The default interpolation is changed to this.

The `ShiftRegistration` tests had been changed because the derivatives were used.


vnmabus:
This is not the way to go. We will rework evaluation and derivation.

PR #228, opened 2020-06-21
Including equal_var parameter in ANOVA

DavidGarciaFer:
Signed-off-by: David García Fernández <david.garcia03@estudiante.uam.es>

- Including homoscedastic case for One-way ANOVA.
- Updating anova synthetic data example, resolving some typos.


examples/plot_oneway_synthetic.py
```diff
@@ -83,16 +81,6 @@
 print("Statistic: {:.3f}".format(stat))
 print("p-value: {:.3f}".format(p_val))
 
-################################################################################
-# In the plot below we can see the simulated trajectories for each mean,
```
vnmabus, line 87: Why is this removed?

PR #267, opened 2020-10-26
update references and change default method

jdtuck:
The DP method is more accurate than DP2. Also added a reference which was the first journal paper on these methods after the arxiv paper

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -22,7 +22,7 @@ Documentation
 The documentation is available at
 `fda.readthedocs.io/en/stable/ <https://fda.readthedocs.io/en/stable/>`_, which
 includes detailed information of the different modules, classes and methods of
-the package, along with several examples showing different funcionalities.
+the package, along with several examples showing different functionalities.
 
 The documentation of the latest version, corresponding with the develop
 version of the package, can be found at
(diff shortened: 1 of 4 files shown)


jdtuck:
so looks like the tests are based on DP2, so they failed. Though a few install failures which looks like travis couldn't install cffi this round

vnmabus:
Yes, they were based on DP2, and the differences between DP and DP2 were high in some cases. I do not know why `cffi` fails to install now.

jdtuck:
I am looking at cffi now, as its not consistent on the travis builds.

jdtuck:
so cffi has to do with appveyor deployment to PyPI, will fix that here shortly.

vnmabus:
BTW, I was trying to port a change that we made to your code to your package (basically, we made the size of the discretization customizable, and we removed `dp_nbhd.h`). We exposed the discretization size parameter in our API (although I am fairly certain that nobody depends on it, so it could be removed if you do not find it worthwhile).

jdtuck:
I would be interested in porting that back, have it on my to do list and have never done it. It would help balance speed/accuracy easier in the DP

jdtuck:
If you port that back, we can go back to DP2, I am fine with that. Also cffi is fixed

PR #273, opened 2020-11-07
plot_with_gradient

mellamansanchez:
New function to plot functions with a gradient of colors

### DIFF
--- a/skfda/representation/_functional_data.py
+++ b/skfda/representation/_functional_data.py
@@ -155,7 +155,7 @@ class FData(ABC, pandas.api.extensions.ExtensionArray):
     @property
     @abstractmethod
     def dim_domain(self):
-        """Return number of dimensions of the :term:`domain`.
+        """Return number of dimensions of the domain.
 
         Returns:
             int: Number of dimensions of the domain.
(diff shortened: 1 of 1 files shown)


vnmabus:
I do not think that a new method is required for the FDataGrid object, at least for now. Also, the implementation is missing.

mellamansanchez:
> I do not think that a new method is required for the FDataGrid object, at least for now. Also, the implementation is missing.



> I do not think that a new method is required for the FDataGrid object, at least for now. Also, the implementation is missing.

I also considered making plot_graph have new variables that can be used in case you wanted to plot the datagrid with a gradient of colors, but I was unsure if we wanted to edit that method.

vnmabus:
> 
> 
> > I do not think that a new method is required for the FDataGrid object, at least for now. Also, the implementation is missing.
> 
> > I do not think that a new method is required for the FDataGrid object, at least for now. Also, the implementation is missing.
> 
> I also considered making plot_graph have new variables that can be used in case you wanted to plot the datagrid with a gradient of colors, but I was unsure if we wanted to edit that method.

Let's discuss that in the next meeting.

PR #280, opened 2020-11-25
Depth examples

pedrorponga:


skfda/ml/classification/DTM_classifier.py
```diff
@@ -0,0 +1,107 @@
+"""Distance to trimmed means (DTM) classification."""
```
vnmabus, line 1: This is already in other PR,

examples/plot_DTM_classification.py
```diff
@@ -0,0 +1,80 @@
+"""Distance to trimmed means (DTM) classification.
```
vnmabus, line 1: I do not think that an separate gallery example for each classification method is THAT useful, unless it has lots of configuration. I think that is better to do a example for each family, such as an example comparing the different depth-based classifiers, and explaining the theory behind each of them and illustrating their differences. Otherwise, this is redundant with the docstring examples.

vnmabus (changes requested):

skfda/ml/classification/DTM_classifier.py
pedrorponga: True

examples/plot_DTM_classification.py
pedrorponga: Makes sense! I will work on a unified example once I have DDClassifier since it will be interesting to show, amongst other things, that MaximumDepthClassifier is just a particular case.

PR #282, opened 2020-11-26
Feature/GitHub actions

pedrorponga:


.travis.yml
```diff
@@ -1,52 +0,0 @@
-language: python
```
vnmabus, line 1: We should keep the Travis config file until all CI is migrated to Github actions (including Windows tests, readme link and removing Travis checks).

pedrorponga: Makes sense

vnmabus (approved):

PR #292, opened 2020-12-04
Linting

pedrorponga:


setup.cfg
```diff
@@ -71,13 +71,17 @@ ignore-decorators = property
 
 strictness = long
 
+# Beautify output and make it more informative
+format = wemake
+show-source = true
+
```
vnmabus, line 77: Wemake says that both options are in the default documentation. Does that mean that this is unnecessary?

pedrorponga: Those two options do change the output when running flake8 from the terminal.

vnmabus (approved):
LGTM.

PR #327, opened 2021-02-24
Plot gr v2

mellamansanchez:


mellamansanchez:
everything seems fine to me

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
vnmabus, line 1: Remove this file.

skfda/exploratory/visualization/representation.py
```diff
@@ -1,32 +1,52 @@
...
         group_colors = colormap(np.arange(n_labels) / (n_labels - 1))
 
     return group_colors
 
 
-def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
+def _get_color_info(
+    fdata: T,
+    group: Union[Sequence[int], None],
+    group_names: Union[Sequence[str], None],
```
vnmabus, line 45: Group names and colors could also be mappings (dicts).

skfda/exploratory/visualization/representation.py
```diff
@@ -1,32 +1,52 @@
...
         colormap = matplotlib.cm.get_cmap()
         group_colors = colormap(np.arange(n_labels) / (n_labels - 1))
 
     return group_colors
 
 
-def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
+def _get_color_info(
+    fdata: T,
+    group: Union[Sequence[int], None],
```
vnmabus, line 44: Group can be a sequence of arbitrary types, like strings.

skfda/exploratory/visualization/representation.py
```diff
@@ -79,249 +104,390 @@ def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
...
+
+    def plot(
+        self,
+        chart: Union[Figure, Axes, None] = None,
+        *,
+        fig: Optional[Figure] = None,
+        ax: Optional[Axes] = None,
+        n_rows: Optional[int] = None,
+        n_cols: Optional[int] = None,
+        n_points: Union[int, Tuple[int, int], None],
```
vnmabus, line 184: I think most `plot` parameters should be moved to `__init__` instead.

skfda/exploratory/visualization/representation.py
```diff
@@ -79,249 +104,390 @@ def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
...
+    Attributes:
+        gradient_list: normalization of the values from gradient color_list
+            that will be used to determine the intensity of the color
+            each function will have.
 
     """
+    def __init__(
+        self,
+        fdata: FData,
+        gradient_color_list: Union[Sequence[float], None] = None,
```
vnmabus, line 138: This parameter is not really a color list. I think it deserves a better name.

vnmabus (changes requested):

skfda/exploratory/visualization/representation.py
mellamansanchez: i have implemented it in the interactive version

skfda/exploratory/visualization/representation.py
```diff
@@ -1,32 +1,61 @@
...
         group_colors = colormap(np.arange(n_labels) / (n_labels - 1))
 
     return group_colors
 
 
-def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
+def _get_color_info(
+    fdata: T,
+    group: Union[Sequence[Any], None],
+    group_names: Union[Sequence[str], Dict[Any, str], None],
```
vnmabus, line 54: Prefer `Mapping` instead of `Dict` if possible. `Mapping` is covariant.

skfda/exploratory/visualization/representation.py
```diff
@@ -79,249 +113,395 @@ def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
...
-        **kwargs: if dim_domain is 1, keyword arguments to be passed to
-            the matplotlib.pyplot.plot function; if dim_domain is 2,
-            keyword arguments to be passed to the
-            matplotlib.pyplot.plot_surface function.
-
-    Returns:
-        fig (figure object): figure object in which the graphs are plotted.
+        fdata: functional data set that we want to plot.
+        gradient_color_list: list of real values used to determine the color
+            in which each of the instances will be plotted. The size
```
vnmabus, line 126: This seems truncated.

skfda/exploratory/visualization/representation.py
```diff
@@ -79,249 +113,395 @@ def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
...
-    """Plot the FDatGrid object graph as hypersurfaces.
+class GraphPlot:
+    """
+    Class used to plot the FDataGrid object graph as hypersurfaces.
 
-    Plots each coordinate separately. If the :term:`domain` is one dimensional,
-    the plots will be curves, and if it is two dimensional, they will be
-    surfaces.
+    A list of variables (probably depths) can be used as an argument to
+    display the functions wtih a gradient of colors.
```
vnmabus, line 121: I would remove this and explain it in the arguments OR explain the different ways in which color can be specified. Also gradients are not only useful for depths, could be useful also for scalar regresion targets, for example.

skfda/exploratory/visualization/representation.py
```diff
@@ -79,249 +113,395 @@ def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
...
+        max_grad: maximum value that the gradient_list can take, it will be
+            used to normalize the gradient_color_list in order to get values
+            thatcan be used in the funcion colormap.__call__(). If not
+            declared it will be initialized to the maximum value of
+            gradient_list
+        min_grad: minimum value that the gradient_list can take, it will be
+            used to normalize the gradient_color_list in order to get values
+            thatcan be used in the funcion colormap.__call__(). If not
+            declared it will be initialized to the minimum value of
+            gradient_list.
```
vnmabus, line 136:
```suggestion
        max_grad: maximum value that the gradient_list can take, it will be
            used to normalize the ``gradient_color_list``. If not
            declared it will be initialized to the maximum value of
            gradient_list
        min_grad: minimum value that the gradient_list can take, it will be
            used to normalize the ``gradient_color_list``. If not
            declared it will be initialized to the minimum value of
            gradient_list.
```

skfda/exploratory/visualization/representation.py
```diff
@@ -79,249 +113,395 @@ def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
...
+
+    Attributes:
+        gradient_list: normalization of the values from gradient color_list
```
vnmabus, line 139: Why should you expose this to the user?

skfda/exploratory/visualization/representation.py
```diff
@@ -79,249 +113,395 @@ def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
...
+            that will be used to determine the intensity of the color
+            each function will have.
 
     """
+    def __init__(
+        self,
+        fdata: FData,
+        gradient_color_list: Union[Sequence[float], None] = None,
```
vnmabus, line 147: If there are only two elements in the union and one of them is `None` prefer `Optional`.

skfda/exploratory/visualization/representation.py
```diff
@@ -79,249 +113,395 @@ def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
...
+    def plot(
+        self,
+        chart: Union[Figure, Axes, None] = None,
+        *,
+        fig: Optional[Figure] = None,
+        ax: Optional[Axes] = None,
+        n_rows: Optional[int] = None,
+        n_cols: Optional[int] = None,
+        n_points: Union[int, Tuple[int, int], None] = None,
+        domain_range: Union[Tuple[int, int], DomainRangeLike, None] = None,
```
vnmabus, line 194:
```suggestion
        domain_range: Optional[DomainRangeLike] = None,
```

skfda/exploratory/visualization/representation.py
```diff
@@ -79,249 +113,395 @@ def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
...
+        group: Union[Sequence[Any], None] = None,
+        group_colors: Union[Sequence[Any], Dict[Any, Any], None] = None,
```
vnmabus, line 196: Same as said above. Use `Any` sparingly.

vnmabus (changes requested):

skfda/exploratory/visualization/representation.py
```diff
@@ -1,32 +1,70 @@
...
         colormap = matplotlib.cm.get_cmap()
         group_colors = colormap(np.arange(n_labels) / (n_labels - 1))
 
     return group_colors
 
 
-def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
+def _get_color_info(
+    fdata: T,
+    group: Union[Sequence[Any], None],
```
vnmabus, line 62: This is a though function to type. Ideally the `@overload` decorator could be used. In that case you would have three overloads:
- One in which group is `None` and then `group_names` and `group_colors` must be also `None`.
- One in which group is `Sequence[int]` and `group_names` is `Union[Sequence[str], Mapping[K, str], None]` and `group_colors` is `Union[Sequence[Any], Mapping[K, Any], None]` (or a better type for the colors if you can find it).
- One in which group is `Sequence[K]` and `group_names` is `Optional[Mapping[K, str]]` and `group_colors` is `Optional[Mapping[K, Any]]`.

The problem with that approach is that currently the implementation is not checked against every overload (see https://github.com/python/mypy/issues/8867), so you woud probably have mypy errors.

Another possibility, that it is arguably better, as it requires less from the arguments is to define and use a [protocol](https://mypy.readthedocs.io/en/stable/protocols.html). You could define a generic protocol `Indexable[K, V]` and then use `Optional[Indexable[K, str]]` for `group_labels`, `Optional[Indexable[K, Any]]` for `group_colors` and `Optional[Sequence[K]]` for `group`.

mellamansanchez: i think i corrected it, i finally went for the overload

skfda/exploratory/visualization/representation.py
```diff
@@ -1,32 +1,80 @@
...
+    def __getitem__(self, __key: K) -> V:
+        pass
 
+
+def _get_label_colors(
+    n_labels: int,
+    group_colors: Union[Sequence[ColorLike], Mapping[K, ColorLike], None],
+) -> np.ndarray:
+    """Get the colors of each label."""
     if group_colors is not None:
```
vnmabus, line 56: Maybe it is also better to reverse the order of the `if-else` in order to use positive conditions.

skfda/exploratory/visualization/representation.py
```diff
@@ -1,32 +1,80 @@
...
 
-def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
+def _get_color_info(
+    fdata: T,
+    group: Optional[Sequence[K]],
+    group_names: Optional[Indexable[K, str]],
+    group_colors: Optional[Indexable[K, ColorLike]],
+    legend: bool,
+    kwargs: Any,
+) -> Tuple[Any, Optional[List[matplotlib.patches.Patch]]]:
```
vnmabus, line 77: Specify the type of the first returned value.

skfda/exploratory/visualization/representation.py
```diff
@@ -79,249 +132,402 @@ def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
...
+        self,
+        chart: Union[Figure, Axes, None] = None,
+        *,
+        fig: Optional[Figure] = None,
+        ax: Optional[Axes] = None,
+        n_rows: Optional[int] = None,
+        n_cols: Optional[int] = None,
+        n_points: Union[int, Tuple[int, int], None] = None,
+        domain_range: Optional[DomainRangeLike] = None,
+        group: Union[Sequence[Any], None] = None,
```
vnmabus, line 210: I do not get why these parameters are in `plot` instead of init. Also, how they interact with the gradients?

skfda/exploratory/visualization/representation.py
```diff
@@ -79,249 +132,402 @@ def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
...
+            gradient_list.
 
     """
 
-    fig, axes = _get_figure_and_axes(chart, fig, axes)
-    fig, axes = _set_figure_layout_for_fdata(fdata, fig, axes, n_rows, n_cols)
+    def __init__(
+        self,
+        fdata: FData,
+        gradient_color_list: Optional[Sequence[float]] = None,
```
vnmabus, line 162: I think we need a better name. Maybe `gradient_values` or even better if we can use a name unrelated with colors (as was done with `group`). Try looking for alternatives to discuss.

skfda/exploratory/visualization/representation.py
```diff
@@ -79,249 +132,402 @@ def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
...
+                self.max_grad = max(self.gradient_color_list)
+            else:
+                self.max_grad = max_grad
+
+            aux_list = [
+                grad_color - self.min_grad
+                for grad_color in self.gradient_color_list
+            ]
+
+            self.gradient_list: Sequence[float] = (
```
vnmabus, line 191: Is the type not inferred by Mypy?

skfda/exploratory/visualization/representation.py
```diff
@@ -79,249 +132,402 @@ def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
...
+        chart: Union[Figure, Axes, None] = None,
+        *,
+        fig: Optional[Figure] = None,
+        ax: Optional[Axes] = None,
+        n_rows: Optional[int] = None,
+        n_cols: Optional[int] = None,
+        n_points: Union[int, Tuple[int, int], None] = None,
+        domain_range: Optional[DomainRangeLike] = None,
+        group: Union[Sequence[Any], None] = None,
+        group_colors: Union[
```
vnmabus, line 211: Use `Indexable` also here.

skfda/exploratory/visualization/representation.py
```diff
@@ -79,249 +132,402 @@ def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
...
-        grid_points = fdata.grid_points
-        evaluated_points = fdata.data_matrix
-
-    if evaluated_points is None:
-        evaluated_points = fdata(
-            grid_points, grid=True)
+    def __init__(
+        self,
+        fdata: FData,
+        grid_points: np.ndarray = None,
```
vnmabus, line 384: `grid_points` usually is a `Optional[GridPointsLike]` object.

skfda/exploratory/visualization/representation.py
```diff
@@ -79,249 +132,402 @@ def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
...
+    def plot(
+        self,
+        chart: Union[Figure, Axes, None] = None,
+        *,
+        fig: Optional[Figure] = None,
+        ax: Optional[Axes] = None,
+        n_rows: Optional[int] = None,
+        n_cols: Optional[int] = None,
+        domain_range: Union[Tuple[int, int], DomainRangeLike, None] = None,
+        group: Union[Sequence[Any], None] = None,
```
vnmabus, line 398: Again, use the types defined above.

vnmabus (changes requested):

skfda/exploratory/visualization/representation.py
mellamansanchez: the thing was that i was going to change that in the next pull, in the branch outliergram_interactive its changed in some

skfda/exploratory/visualization/representation.py
```diff
@@ -79,249 +134,394 @@ def _get_color_info(fdata, group, group_names, group_colors, legend, kwargs):
...
+            sample_colors = [None] * self.fdata.n_samples
+            for m in range(self.fdata.n_samples):
+                sample_colors[m] = colormap(self.gradient_list[m])
 
-    if fdata.dim_domain == 1:
+        self.sample_colors = sample_colors
 
-        if n_points is None:
-            n_points = constants.N_POINTS_UNIDIMENSIONAL_PLOT_MESH
+        color_dict: Mapping[str, Any] = {}
```
vnmabus, line 309: Why `Any`?

vnmabus (changes requested):

vnmabus (approved):
LGTM

PR #333, opened 2021-03-14
Refactor registration scorers

vnmabus:
- Added typing and refactor code according to our style guide.
- All but one scorer do not longer require the discretization of functions, relying instead on the L2 distance.
- Fixed some errors where the L2 distance was not properly squared to match the formula.

@pablomm I know you are not longer as involved with the project as before, but as you are the original author of this module I appreciate any suggestions that you can make.

### DIFF
--- a/skfda/preprocessing/registration/validation.py
+++ b/skfda/preprocessing/registration/validation.py
@@ -1,14 +1,17 @@
 """Methods and classes for validation of the registration procedures"""
+from __future__ import annotations
 
-from typing import NamedTuple, Optional
+from abc import ABC, abstractmethod
+from typing import NamedTuple, Optional, Union
 
 import numpy as np
 
 from ..._utils import _to_grid, check_is_univariate
 from ...representation import FData
+from .base import RegistrationTransformer
 
 
-class RegistrationScorer():
+class RegistrationScorer(ABC):
     """Cross validation scoring for registration procedures.
 
     It calculates the score of a registration procedure, used to perform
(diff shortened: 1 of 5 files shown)


pablomm:
Hi @vnmabus , thanks for taking me into account. I'll check it out tonight.

skfda/preprocessing/registration/validation.py
```diff
@@ -111,15 +139,14 @@ class AmplitudePhaseDecomposition(RegistrationScorer):
...
+        C_R \int \overline{y}^2(t) dt - \int \overline{x}^2(t) dt
 
     where the constant :math:`C_R` is defined as
 
     .. math::
 
-        C_R = 1 + \frac{\frac{1}{N}\sum_{i}^{N}\int [Dh_i(t)-\overline{Dh}(t)]
-        [ y_i^2(t)- \overline{y^2}(t) ]dt}
-        {\frac{1}{N} \sum_{i}^{N} \int y_i^2(t)dt}
+        C_R = \frac{\frac{1}{N}\sum_{i=1}^{N}\int[x_i(t)-\overline x(t)]^2dt
```
pablomm, line 148: Are the Cr formulas equivalent or was the previous one wrong?
If they are equivalent, it is a great simplification avoiding the use of h(t).

pablomm (approved):
Everything I have seen seems to be correct. 

I have not been able to see how the simplification of the Cr formula is done, but I assume it is correct.

p.d: I couldn't run the code due to the compilation problems.

skfda/preprocessing/registration/validation.py
vnmabus: The previous formula is derived from this one in the original paper. I think that they derive it for theoretical reasons (in order to know how the warpings influence the decomposition) but for the implementation I think the original one is better as you do not need the warpings.

vnmabus:
Did you check the missing squares in some formulas? They use the l2_norm squared, as they not compute the square root.

skfda/preprocessing/registration/validation.py
pablomm: I didn't see that part when I implemented it. 👍

pablomm:
> Did you check the missing squares in some formulas? They use the l2_norm squared, as they not compute the square root.

Yes, I review the formulas of amplitude-phase, least squares, and Sobolev scorers taking that into account. If I haven't skipped anything, all square roots of the norm cancel with the squares.

PR #344, opened 2021-04-14
Base plot

mellamansanchez:


skfda/exploratory/visualization/_baseplot.py
```diff
@@ -0,0 +1,94 @@
...
+        Abstract method used to plot the object and its data.
+
+        Returns:
+            Figure: figure object in which the displays and
+                widgets will be plotted.
+        """
+        pass
+
+    @abstractmethod
+    def num_instances(self) -> int:
```
vnmabus, line 53: Why not `n_samples`?

skfda/exploratory/visualization/_baseplot.py
```diff
@@ -0,0 +1,94 @@
...
+        """Get the number of instances that will be used for interactivity."""
+        pass
+
+    @abstractmethod
+    def set_figure_and_axes(
```
vnmabus, line 58: Why is this part of the API?

skfda/exploratory/visualization/_baseplot.py
mellamansanchez: because depending on the object it can change, for example parametricplot can have two fdata

skfda/exploratory/visualization/_baseplot.py
mellamansanchez: i added it to have in a single method all the initialzation of the axes and figure, it is overriden with the different cases in each class

skfda/exploratory/visualization/_baseplot.py
vnmabus: And how the interactivity works in that case?

mellamansanchez: as it uses the fdata final there is no problem, the thing is that when we use a baseplot object in multiple display
i need a generic way for n_samples

mellamansanchez:
as the methods are used in multiple display i didnt put underscore, i was getting a warning

skfda/exploratory/visualization/_baseplot.py
```diff
@@ -0,0 +1,82 @@
...
+            the visualization with interactivity and widgets.
+        fig: Figure over with the graphs are plotted.
+        axes: Sequence of axes where the graphs are plotted.
+    """
+
+    @abstractmethod
+    def __init__(
+        self,
+    ) -> None:
+        self.artists: List[Artist] = []
```
vnmabus, line 35: Shouldn't this be an array of length `self.n_samples`?

skfda/exploratory/visualization/_baseplot.py
```diff
@@ -0,0 +1,82 @@
...
+        self.fig: Optional[Figure] = None
```
vnmabus, line 36: Shouldn't `fig` and `axes` be init parameters?

skfda/exploratory/visualization/_baseplot.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        pass
+
+    @abstractmethod
+    def n_samples(self) -> int:
+        """Get the number of instances that will be used for interactivity."""
+        pass
+
+    @abstractmethod
+    def set_figure_and_axes(
```
vnmabus, line 58: So, which differences should be in subclasses? Is this the right abstraction?

mellamansanchez: it was a way to call all the initialization axes and fig methods from utils the same way, some times we initialize it with some parameters other times with others, that is the reason for this method

skfda/exploratory/visualization/_baseplot.py
mellamansanchez: they can be init parameters, but they wont be used until set figure and axes is called

skfda/exploratory/visualization/_baseplot.py
vnmabus: I think you can create an array of fixed length.

skfda/exploratory/visualization/_baseplot.py
vnmabus: Is there any plot where they won't be passed on init?

skfda/exploratory/visualization/_baseplot.py
vnmabus: Can you be more concrete?

skfda/exploratory/visualization/_baseplot.py
mellamansanchez: do you mean a numpy array? or just a normal one

skfda/exploratory/visualization/_baseplot.py
mellamansanchez: for example the boxplot uses _get_figure_and_axes, _set_figure_layout_for_fdata to initialize the axes, while FPCA only uses _get_figure_and_axes and does a condition. In this way i encapsulate in every file the functionality to initialize axes for a clearer understanding

skfda/exploratory/visualization/_baseplot.py
mellamansanchez: no, it will be in everyone

skfda/exploratory/visualization/_baseplot.py
vnmabus: Numpy array.

skfda/exploratory/visualization/_baseplot.py
vnmabus: Where is this called? In plot?

mellamansanchez: we always initialize the axes in init, you can see in the outliergram interactive branch

vnmabus: Then why do we need this method? Couldn't this be done in init (which is different for each plot type) and avoid exposing another public method?

mellamansanchez: ok, so we do it in init, it is unnecessary

vnmabus (approved):

mellamansanchez:
carlos, a change has been made in the library so i had to merge, please accept again the pr

vnmabus:
> carlos, a change has been made in the library so i had to merge, please accept again the pr

I think it is not necessary to approve it again, just let the tests pass.

PR #346, opened 2021-04-22
ddplot

mellamansanchez:


skfda/exploratory/visualization/_ddplot.py
```diff
@@ -127,4 +121,35 @@ def plot(
...
         )
 
-        return fig
+        return self.fig
+
+    def n_samples(self) -> int:
+        """Get the number of instances that will be used for interactivity."""
+        return self.fdata.n_samples
+
+    def set_figure_and_axes(
```
vnmabus, line 130: This should not be public.

skfda/exploratory/visualization/_ddplot.py
```diff
@@ -72,41 +81,26 @@ def plot(
...
-            ax: axis where the graphs are plotted. If None, see param fig.
-            kwargs: if dim_domain is 1, keyword arguments to be passed to the
-                matplotlib.pyplot.plot function; if dim_domain is 2, keyword
-                arguments to be passed to the matplotlib.pyplot.plot_surface
-                function.
         Returns:
             fig (figure object): figure object in which the depths will be
             scattered.
         """
+        self.artists = np.array([])
```
vnmabus, line 88: Create the array with the final dimension.

skfda/exploratory/visualization/_ddplot.py
```diff
@@ -72,41 +81,26 @@ def plot(
...
+        for d1, d2 in zip(self.depth_dist1, self.depth_dist2):
+            self.artists = np.append(self.artists, ax.scatter(
+                d1,
+                d2,
+                picker=2,
+            ))
 
         # Set labels of graph
-        fig.suptitle("DDPlot")
+        ax.set_title("DDPlot")
```
vnmabus, line 103: Probably better to use the dataset name. I am not opposed to adding an optional `name` parameter for `BasePlot` though.

vnmabus (changes requested):

skfda/exploratory/visualization/_ddplot.py
mellamansanchez: i just realised again that this is called from multiple display, should i call it _set_figure_and_axes and import it? the other day it was giving me an error, with utils methods it works but as this one is a class method i dont know

vnmabus: Why is this called from multiple display?

mellamansanchez: because the baseplot is already created, we do a copy of it and then modify where it is displayed

skfda/exploratory/visualization/_ddplot.py
```diff
@@ -46,7 +56,11 @@ def __init__(
         dist1: T,
         dist2: T,
         depth_method: Depth[T],
+        chart: Union[Figure, Axes, None] = None,
+        fig: Optional[Figure] = None,
+        axes: Optional[Axes] = None,
```
vnmabus, line 61: I still think that the chart parameter should be removed... but if it is not it should be a positional parameter (and most parameters should NOT be positional).


```suggestion
        dist1: T,
        dist2: T,
        chart: Union[Figure, Axes, None] = None,
        *,
        depth_method: Depth[T],
        fig: Optional[Figure] = None,
        axes: Optional[Axes] = None,
```

vnmabus (changes requested):

skfda/exploratory/visualization/_ddplot.py
mellamansanchez: ok, i will do this in every visualization method

skfda/exploratory/visualization/_ddplot.py
```diff
@@ -72,41 +83,27 @@ def plot(
...
-        ax.scatter(
-            self.depth_dist1,
-            self.depth_dist2,
-            **kwargs,
-        )
+        for i in range(len(self.depth_dist1)):
+            self.artists[i] = ax.scatter(
+                self.depth_dist1[i],
+                self.depth_dist1[i],
+                picker=2,
```
vnmabus, line 101: Why 2? Shouldn't be a boolean?

vnmabus (changes requested):

skfda/exploratory/visualization/_ddplot.py
mellamansanchez: it can be a boolean or the tolerance distance that activates the point when clicked

skfda/exploratory/visualization/_ddplot.py
```diff
@@ -72,41 +83,27 @@ def plot(
...
 
-        ax.scatter(
-            self.depth_dist1,
-            self.depth_dist2,
-            **kwargs,
-        )
+        for i in range(len(self.depth_dist1)):
+            self.artists[i] = ax.scatter(
+                self.depth_dist1[i],
+                self.depth_dist1[i],
```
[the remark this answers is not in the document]
mellamansanchez: changed it by mistake in last commit mb

skfda/exploratory/visualization/_ddplot.py
vnmabus: Use `pickradius` for that: it is more clear.

vnmabus (approved):

PR #347, opened 2021-04-22
MagnitudeShapePlot

mellamansanchez:


skfda/exploratory/visualization/_magnitude_shape_plot.py
```diff
@@ -158,14 +163,23 @@ class MagnitudeShapePlot:
 
     """
 
-    def __init__(self, fdatagrid, **kwargs):
+    def __init__(
+        self,
+        fdatagrid: FDataGrid,
+        *,
+        chart: Union[Figure, Axes, None] = None,
```
vnmabus, line 170: Again, if we do not remove chart, it should be positional.

skfda/exploratory/visualization/_magnitude_shape_plot.py
```diff
@@ -270,34 +293,48 @@ def outliercol(self, value):
...
+        self.axes[0].set_xlabel(self.xlabel)
+        self.axes[0].set_ylabel(self.ylabel)
+        self.axes[0].set_title(self.title)
+
+        return self.fig
+
+    def n_samples(self) -> int:
+        return self.fdatagrid.n_samples
+
+    def set_figure_and_axes(
```
vnmabus, line 328: Make it protected.

vnmabus (changes requested):
Notice that this plot is also used in examples and the documentation. Make sure that there are no errors when generating the docs.

mellamansanchez:
> Notice that this plot is also used in examples and the documentation. Make sure that there are no errors when generating the docs.

I have been testing it and it appears that it calls by default _repr_svg_ instead __repr__, which is the one called in the example, but if i call repr it return a str instead of the example, if i use print it gets differences anyway...

vnmabus:
> > Notice that this plot is also used in examples and the documentation. Make sure that there are no errors when generating the docs.
> 
> I have been testing it and it appears that it calls by default _repr_svg_ instead **repr**, which is the one called in the example, but if i call repr it return a str instead of the example, if i use print it gets differences anyway...

I just say that you should generate the docs (with a previous make clean) in order to see if one of them fails.

mellamansanchez:
> > > Notice that this plot is also used in examples and the documentation. Make sure that there are no errors when generating the docs.
> > 
> > 
> > I have been testing it and it appears that it calls by default _repr_svg_ instead **repr**, which is the one called in the example, but if i call repr it return a str instead of the example, if i use print it gets differences anyway...
> 
> I just say that you should generate the docs (with a previous make clean) in order to see if one of them fails.

it is indeed failing at the tests, but i dont know if modifying the example and calling directly __repr__() and then printing it would be acceptable

vnmabus:
Which error do you have?

mellamansanchez:
> Which error do you have?

the example of the magnitude shape plot isnot giving the same result as now we have a different repr method (repr_svg) if i call
__repr__ it returns a string and its not equal even if i change it or print it

skfda/exploratory/visualization/_magnitude_shape_plot.py
```diff
@@ -5,19 +5,24 @@
...
+from matplotlib.axes import Axes
+from matplotlib.figure import Figure
 
+from ... import FDataGrid
 from ..outliers import DirectionalOutlierDetector
-from ._utils import _figure_to_svg, _get_figure_and_axes, _set_figure_layout
+from ._baseplot import BasePlot
+from ._utils import _get_figure_and_axes, _set_figure_layout
 
 __author__ = "Amanda Hernando Bernabé"
```
vnmabus, line 21: Remove author/email. Several authors have modified this file.

skfda/exploratory/visualization/_magnitude_shape_plot.py
```diff
@@ -158,14 +163,23 @@ class MagnitudeShapePlot:
...
 
-    def __init__(self, fdatagrid, **kwargs):
+    def __init__(
+        self,
+        fdatagrid: FDataGrid,
+        chart: Union[Figure, Axes, None] = None,
+        *,
+        fig: Optional[Figure] = None,
+        axes: Optional[Sequence[Axes]] = None,
+        **kwargs,
```
vnmabus, line 173: Are they used?

skfda/exploratory/visualization/_magnitude_shape_plot.py
```diff
@@ -270,34 +293,49 @@ def outliercol(self, value):
...
-
         Returns:
             fig (figure object): figure object in which the graph is plotted.
 
         """
 
-        fig, axes = _get_figure_and_axes(chart, fig, axes)
-        fig, axes = _set_figure_layout(fig, axes)
-
+        self.artists = np.array([])
```
vnmabus, line 304: Again, reserve the full size array.

vnmabus (changes requested):

skfda/exploratory/visualization/_magnitude_shape_plot.py
mellamansanchez: kwargs is used by the DirectionalOutlierDetector

skfda/exploratory/visualization/_magnitude_shape_plot.py
```diff
@@ -207,11 +228,13 @@ def __init__(self, fdatagrid, **kwargs):
         self._outliers = outliers
         self._colormap = plt.cm.get_cmap('seismic')
         self._color = 0.2
-        self._outliercol = 0.8,
+        self._outliercol = 0.8
         self.xlabel = 'MO'
         self.ylabel = 'VO'
         self.title = 'MS-Plot'
```
vnmabus, line 234: There are nice ideas in this class, that can be used as part of the API in general. For example, having a `title` property, initialized to the dataset name (and not to the type of plot) but that can be modified by the user to change the plot title (or remove it). Same with `xlabel` and `ylabel`. Have you though on this?

mellamansanchez: I was not planning on adding extra features to the current versions as I wanted to arrive to the ordinary handing over date of the project

vnmabus: Ok, add it to future work and create an issue if you think that this design is desirable. In this plot, I would change the title to the dataset name.

mellamansanchez: Ok perfect

mellamansanchez: i have solved the issues with the example, i think this file is now okay

skfda/exploratory/visualization/_magnitude_shape_plot.py
```diff
@@ -109,7 +107,7 @@ class MagnitudeShapePlot(BasePlot):
         ...                [-0.5, -0.5, -0.5, -1, -1, -1]]
         >>> grid_points = [ 0., 2., 4., 6., 8., 10.]
         >>> fd = skfda.FDataGrid(data_matrix, grid_points)
-        >>> MagnitudeShapePlot(fd)
```
vnmabus, line 107: Why didn't this work.

mellamansanchez: it didnt work because know it was calling _repr_svg_() instead of __repr__()

vnmabus: That should not be possible. Have you debugged it to know where `_repr_svg_` is called?

mellamansanchez: my bad, it actually works gonna upload it

skfda/exploratory/visualization/_magnitude_shape_plot.py
```diff
@@ -131,20 +134,24 @@ class MagnitudeShapePlot:
                         [-1. ],
                         [-1. ],
                         [-1. ]]]),
-                grid_points=(array([ 0.,  2.,  4.,  6.,  8., 10.]),),
+                grid_points=(array([  0.,   2.,   4.,   6.,   8.,  10.]),),
                 domain_range=((0.0, 10.0),),
-                ...),
+                dataset_name=None,
```
vnmabus, line 139: Why did you add everything?

skfda/exploratory/visualization/_magnitude_shape_plot.py
```diff
@@ -131,20 +134,24 @@ class MagnitudeShapePlot:
...
+                extrapolation=None,
+                interpolation=SplineInterpolation(interpolation_order=1, smoothness_parameter=0, monotone=False)),
             multivariate_depth=None,
             pointwise_weights=None,
             alpha=0.993,
             points=array([[ 1.66666667,  0.12777778],
-                          [ 0.        ,  0.        ],
-                          [-0.8       ,  0.17666667],
-                          [-1.74444444,  0.94395062]]),
+                [ 0.        ,  0.        ],
```
vnmabus, line 148: Why did you remove indentation?

vnmabus (changes requested):

skfda/exploratory/visualization/_magnitude_shape_plot.py
mellamansanchez: because due to the changes it was now printing it, and if i didnt put it would be me an error as the expected was different to the output

skfda/exploratory/visualization/_magnitude_shape_plot.py
mellamansanchez: because it had changed, if i didnt touch it would be different

skfda/exploratory/visualization/_magnitude_shape_plot.py
vnmabus: It also printed it before the changes. The ellipsis should match anything, though.

skfda/exploratory/visualization/_magnitude_shape_plot.py
vnmabus: Whitespace should be ignored in the comparisons, if the setup.cfg is correct.

skfda/exploratory/visualization/_magnitude_shape_plot.py
mellamansanchez: ok, ellipsis is working mb

skfda/exploratory/visualization/_magnitude_shape_plot.py
mellamansanchez: ok, just rrevised it and corrected it, the params are correct in setup.cfg too

vnmabus (approved):
LGTM

PR #389, opened 2021-11-04
Example fpca.inverse_transform to detect outliers

Clej:
I propose an example of using ``inverse_transform`` in ``FPCA`` to detect outliers from the reconstruction error (RE) between the input and the _recovered_ input from the eigenspace.
In other words, given a fitted fpca, a threshold and a test sample ``x`` the decision rule is:
```javascript
x_hat  = fpca.inverse_transform(fpca.transform(x))
RE = lp_distance(x, x_hat, p=2) / lp_norm(x, p=2)
if RE>= threshold:
   print('x is an outlier')
```

In the example, we use synthetic data generated from centered gaussian processes. The nonoutliers are generated with a Gaussian kernel and the outliers are generated with an exponential kernel. There are only nontouliers in the training dataset (used to fit the FPCA and set the threshold). There are both outliers and nonoutliers in the test samples.

I tried to make my explanations as concise and intuitive as possible.

The script generates two figures: (i) the dataset and (ii) the distribution of the REs.
(i)
dataset_fpca_outl

(ii)
RE_density_fpca_outl

Finally, we output the truely detected outliers and nonoutliers.

### DIFF
--- a/examples/plot_fpca_inverse_transform_outl_detection.py
+++ b/examples/plot_fpca_inverse_transform_outl_detection.py
@@ -38,9 +38,9 @@ from skfda.misc.metrics import lp_distance, lp_norm
 # - We generate a clean training dataset (not supposed to contain outliers)
 # and fit an FPCA with 'q' components on it.
 # - We also generate a test set containing
-# both nonoutliers samples and outliers.
+# both nonoutliers and outliers samples.
 # - Then, we fit an FPCA(n_components=q)
-# and compute the vector of principal components scores
+# and compute the vectors of principal components scores
 # of train and test samples.
 # - We project back the vectors of principal components scores,
 # with the inverse_transform method, to the input (training data space).
@@ -157,7 +157,11 @@ err_test = lp_distance(
 ###########################################################################
 # We plot the density of the REs,
 # both unconditionnaly (grey and blue) and conditionnaly (orange and red),
-# to the rule error >= threshold.
+# to the rule if error >= threshold then it is an outlier.
+# The threshold is computed from RE of the training samples as
+# the quantile of probability 0.99.
+# In ohter words, a sample whose RE is higher than the threshold is unlikely
+# approximated as a training sample with (low) probability 0.01.
 x_density = np.linspace(0., 1.6, num=10**3)
 density_train_err = gaussian_kde(err_train)
 density_test_err = gaussian_kde(err_test)
--- a/examples/plot_fpca_inverse_transform_outl_detection.py
+++ b/examples/plot_fpca_inverse_transform_outl_detection.py
@@ -198,6 +198,14 @@ plt.title('Densities of reconstruction errors with {} components'.format(q))
 plt.legend()
 plt.show()
 
+####################################################################
+# We can check that the outliers are all detected with this method,
+# with no false positive (wrongly) in the test set.
+print('Flagged outliers: \n',
+    test_set[err_test>=err_thresh].sample_names)
+print('Flagged nonoutliers: \n',
+    test_set[err_test<err_thresh].sample_names)
+
 ##############################################################################
 # We observe that the distribution of the training samples (grey) REs
 # is unimodal and quite concentrated toward 0. This means that
--- a/examples/plot_fpca_inverse_transform_outl_detection.py
+++ b/examples/plot_fpca_inverse_transform_outl_detection.py
@@ -1,0 +1,220 @@
+"""
+Outlier detection with FPCA
+===========================
+
+Example of using the inverse_transform method
+in the FPCA class to detect outlier(s) from
+the reconstruction (truncation) error.
+
+In this example, we illustrate the utility of the inverse_transform method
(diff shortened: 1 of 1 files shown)


examples/plot_fpca_inverse_transform_outl_detection.py
```diff
@@ -0,0 +1,231 @@
...
+# density estimate of the error of test samples flagged as outliers
+plt.plot(x_density, density_test_err_outl(x_density),
+         label='Error test>= thresh (outliers)', color='C1')
+
+# density estimate of the error of test samples flagged as nonoutliers
+plt.plot(x_density, density_test_err_inli(x_density),
+         label='Error test< thresh (nonoutliers)', color='C3')
+
+plt.xlabel('Relative L2-norm reconstruction errors')
+plt.ylabel('Density (unormalized)')
```
vnmabus, line 195: Typo:
```suggestion
plt.ylabel('Density (unnormalized)')
```

examples/plot_fpca_inverse_transform_outl_detection.py
```diff
@@ -0,0 +1,231 @@
...
+plt.title('Densities of reconstruction errors with {} components'.format(q))
+plt.legend()
+plt.show()
+
+####################################################################
+# We can check that the outliers are all detected with this method,
+# with no false positive (wrongly) in the test set.
+print('Flagged outliers: \n',
```
vnmabus, line 203: I think I prefer to call print twice than to pass two arguments and embed a newline.

examples/plot_fpca_inverse_transform_outl_detection.py
```diff
@@ -0,0 +1,231 @@
...
+      test_set[err_test >= err_thresh].sample_names)
+print('Flagged nonoutliers: \n',
+      test_set[err_test < err_thresh].sample_names)
+
+##############################################################################
+# We observe that the distribution of the training samples (grey) REs
+# is unimodal and quite concentrated toward 0. This means that
+# the training samples are well recovered with 5 FPCs if we allow
+# an error rate around 0.4. On the countrary, the distribution of the
```
vnmabus, line 212: Typo:
```suggestion
# an error rate around 0.4. On the contrary, the distribution of the
```

examples/plot_fpca_inverse_transform_outl_detection.py
```diff
@@ -0,0 +1,231 @@
...
+# test samples (blue) is bimodal
+# with equivalent an magnitude order for each mode,
```
vnmabus, line 214: I do not get the meaning of the sentence as written. Also, another typo:
```suggestion
# with equivalent a magnitude order for each mode,
```

examples/plot_fpca_inverse_transform_outl_detection.py
```diff
@@ -0,0 +1,231 @@
...
+Example of using the inverse_transform method
+in the FPCA class to detect outlier(s) from
+the reconstruction (truncation) error.
+
+In this example, we illustrate the utility of the inverse_transform method
+of the FPCA class to perform functional outlier detection.
+Roughly speaking, an outlier is a sample
+which is not representative of the dataset
+or different enough compared to a large part of the samples.
+The intuition is the following: if the eigen basis,
```
vnmabus, line 14: I think eigenbasis is one word:
```suggestion
The intuition is the following: if the eigenbasis,
```

vnmabus (changes requested):

examples/plot_fpca_inverse_transform_outl_detection.py
```diff
@@ -0,0 +1,231 @@
...
+
+n_train = 10**3
+train_set = make_gaussian_process(
+    n_samples=n_train,
+    n_features=grid_size,
+    start=0., stop=25.,
+    cov=cov_clean,
+    random_state=20
+)
+train_set.sample_names = ['train_' + str(i) for i in range(n_train)]
```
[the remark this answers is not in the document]
Clej: you're right, no need. Labels are enough

examples/plot_fpca_inverse_transform_outl_detection.py
Clej: I've changed it to 'where the two modes seem to be similar'

examples/plot_fpca_inverse_transform_outl_detection.py
```diff
@@ -0,0 +1,231 @@
...
+    'test_outl_' + str(i) for i in range(test_set_outlier.n_samples)]
+
+test_set = test_set_clean.concatenate(test_set_outlier)
+
+#############################
+# We plot the whole dataset.
+whole_data = train_set.concatenate(test_set)
+
+labels = []
+for i in whole_data.sample_names:
```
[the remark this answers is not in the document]
Clej: Actually no need to set sample names since we use them as surrogate for label prediction when assessing ``err_test >= err_thresh``.

vnmabus (approved):
LGTM.

PR #390, opened 2021-11-08
Support for dataframes

rafa9811:
I'm not sure about the correct place for the new mixin class. I placed it in linear_regression file, for the moment.


skfda/ml/regression/_linear_regression.py
```diff
@@ -49,10 +50,44 @@
 ]
 
 
+class DataFrameMixin:
```
vnmabus, line 53: Why is a mixin needed if a simple function can do the job?

skfda/ml/regression/_linear_regression.py
```diff
@@ -49,10 +50,44 @@
...
+        covariates and the second, a FDataBasis object with functional data.
+
+        Args:
+            - X: pandas DataFrame to convert.
+
+        Returns:
+            - list with two elements: first of all, a list with mv
+            covariates and the second, a FDataBasis object with functional data.
+        """
+        fdb = FDataBasis.concatenate(*X.iloc[:, 1].tolist())
```
vnmabus, line 67: I think we want to convert to a list with one variable (multivariate or functional, whatever the case) per column. Don't try to concatenate things or apply fancy logic.

vnmabus (changes requested):

skfda/ml/regression/_linear_regression.py
```diff
@@ -49,10 +50,44 @@
...
+
+
 class LinearRegression(
-    BaseEstimator,  # type: ignore
+    BaseEstimator,   # type: ignore
     RegressorMixin,  # type: ignore
+    DataFrameMixin,  # type: ignore
 ):
+
+    """
```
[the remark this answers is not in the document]
vnmabus: Why did you move the docstring?

skfda/ml/regression/_linear_regression.py
```diff
@@ -49,10 +51,36 @@
...
     BaseEstimator,  # type: ignore
-    RegressorMixin,  # type: ignore
+    RegressorMixin, # type: ignore
+    DataFrameMixin, # type: ignore
 ):
+    """
+        .. deprecated:: 0.8
+        Use covariate parameters of type pandas.FDataFrame in methods fit, predict.
+    """
+    warnings.warn(
```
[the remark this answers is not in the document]
vnmabus: Are you showing a warning even when the right parameters are passed?

skfda/ml/regression/_linear_regression.py
```diff
@@ -50,9 +51,24 @@
...
+
+    """
+    .. deprecated:: 0.8
+    Use covariate parameters of type pandas.FDataFrame in methods
+    fit, predict.
+
+    """
+
+    warnings.warn(
+        "Usage of arguments of type FData, ndarray or a sequence \
```
vnmabus, line 66: The only deprecated case is passing a sequence. FData and ndarray are still fine.

skfda/ml/regression/_linear_regression.py
```diff
@@ -333,3 +380,17 @@ def _argcheck_X_y(
                 )
 
         return new_X, y, sample_weight, coef_info
+    
+    def dataframe_conversion(self, X: pd.DataFrame) -> List:
```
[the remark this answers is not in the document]
vnmabus: This method should not be public.

skfda/ml/regression/_linear_regression.py
```diff
@@ -333,3 +380,17 @@ def _argcheck_X_y(
...
+        """Convert DataFrames to a list with two elements: first of all, a list with mv
```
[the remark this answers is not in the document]
vnmabus: That is not right. The dataframe may have several ndarrays or FData objects, and they can be interleaved. Just extract them to a list in the same order.

vnmabus (changes requested):

skfda/ml/regression/_linear_regression.py
```diff
@@ -150,6 +158,55 @@ class LinearRegression(
...
+
+        First example:
+
+        >>> x_basis = Monomial(n_basis=3)
+        >>> cov_list = [ ( FDataBasis(x_basis, [0, 0, 1]) ),
+        ...              ( FDataBasis(x_basis, [0, 1, 0]) ),
+        ...              ( FDataBasis(x_basis, [0, 1, 1]) ),
+        ...              ( FDataBasis(x_basis, [1, 0, 1]) )]
+        >>> y = [2, 3, 4, 5]
+        >>> df = pd.DataFrame(cov_list)
```
vnmabus, line 171: Add `pd` import to the example.

vnmabus, line 171: This is wrong. Each covariate should be a different column, and the rows should be different samples.

vnmabus (changes requested):

vnmabus (changes requested):

skfda/ml/regression/_linear_regression.py
```diff
@@ -65,6 +66,10 @@ class LinearRegression(
     where the covariates can be either multivariate or functional and the
     response is multivariate.
 
+    .. deprecated:: 0.8.
+        Use covariate parameters of type pandas.DataFrame in methods
```
vnmabus, line 70: That is NOT what is deprecated.

vnmabus (changes requested):

skfda/ml/regression/_linear_regression.py
```diff
@@ -330,3 +400,49 @@ def _argcheck_X_y(
                 )
 
         return new_X, y, sample_weight, coef_info
+
+    def __dataframe_conversion(
```
vnmabus, line 404: Why double underscore?

skfda/ml/regression/_linear_regression.py
```diff
@@ -330,3 +400,49 @@ def _argcheck_X_y(
...
+
+        Returns:
+            List: first of all, a list with mv covariates
+            and the second, a list of FDataBasis object with functional data
+        """
+        mv_list = []
+        fd_list = []
+        final = []
+
+        for obs in X.values.tolist():
```
vnmabus, line 424: You need to iterate over the covariates directly, appending them to a list. I think you can just do
```python
return [v.values for k, v in X.items()]
```
which is both easier and more efficient.

vnmabus (changes requested):

skfda/ml/regression/_linear_regression.py
rafa9811: This way does not support columns of single numbers instead of array. We talked about allow users to use dataframes in two ways:
1) 
`
multivariate1 = [0, 2, 1, 3, 4, 2, 3]
multivariate2 = [0, 7, 7, 9, 16, 14, 5]
cov_dict = {"fd": x_fd, "mult1": multivariate1, "mult2": multivariate2}
df = pd.DataFrame(cov_dict)
`
2) 
`
multivariate = [[0, 0], [2, 7], [1, 7], [3, 9], [4, 16], [2, 14], [3, 5]]
cov_dict = {"fd": x_fd, "mult": multivariate}
df = pd.DataFrame(cov_dict)
`

vnmabus: Why does it not support that?

skfda/ml/regression/_linear_regression.py
```diff
@@ -150,6 +157,58 @@ class LinearRegression(
         >>> linear.predict([x, x_fd])
         array([ 11.,  10.,  12.,   6.,  10.,  13.])
 
+        Funcionality with pandas Dataframe.
+
+        First example:
+
+        >>> x_basis = Monomial(n_basis=3)
+        >>> x_fd = FDataBasis(x_basis, [[0, 0, 1],
```
vnmabus, line 165: Format docstring code using our style guide as well.

vnmabus (approved):

PR #393, opened 2021-11-13
Logistic regression

dSerna4:


vnmabus (changes requested):

_logistic_regression.py
```diff
@@ -0,0 +1,193 @@
...
+        n_features = len(X.grid_points[0])
+
+        selected_indexes = np.zeros(self.p, dtype=np.intc)
+
+        mvlr = mvLogisticRegression()  # multivariate logistic regression
+
+        x_mv = np.zeros((n_samples, self.p))
+        LL = np.zeros(n_features)
+        for q in range(self.p):
+            for t in range(n_features):
```
vnmabus, line 99: This is very slow, right? Couldn't `LL` be considered a continuous function of `t`, and then use optimization techniques for functions, so that we don't need to evaluate it for every `t`? Is that considered anywhere in the literature?

dSerna4: I thought about using optimization techniques for `LL` as well, but I discarded that option because in this algorithm you have to fit the parameters for each `t`.

vnmabus: I am not sure why it should be a problem. My proposal is to consider that the thing computed in the inner loop is a continuous function of `t` (please, check that it is indeed continuous). If that is the case, computing it for values of `t` that are nearby is wasteful, as the value of `LL` will not change much.

Thus, we could use any of the methods in `https://docs.scipy.org/doc/scipy/reference/optimize.html#global-optimization` to optimize that function. In addition, this extends the applicability of this method to basis representation, and multiple domain dimensions as it only requires function evaluation.

I think we should discuss this in the next meeting.

dSerna4: Okay, I think I understand now. 

> I think we should discuss this in the next meeting.
Perfect.

dSerna4: Okay, I think I understand now.

> I think we should discuss this in the next meeting.

Perfect.

_logistic_regression.py
```diff
@@ -0,0 +1,193 @@
...
+
+                x_mv[:, q] = X.data_matrix[:, t, 0]
+                mvlr.fit(x_mv[:, :self.p + 1], y_ind)
```
vnmabus, line 102: Shouldn't this end in `q+1` instead?

dSerna4: yes, of course. It's already fixed. The doctest didn't catch the error because in it I had set p=2. That have made me worried about the usefulness of the doctest, so I've improved it.

skfda/ml/classification/_logistic_regression.py
```diff
@@ -0,0 +1,193 @@
...
+        LL = np.zeros(n_features)
+        for q in range(self.p):
+            for t in range(n_features):
+
+                x_mv[:, q] = X.data_matrix[:, t, 0]
+                mvlr.fit(x_mv[:, :q + 1], y_ind)
+
+                # log-likelihood function at t
+                log_probs = mvlr.predict_log_proba(x_mv[:, :q + 1])
+                log_probs = np.array(
```
vnmabus, line 106: I think this can be vectorized.

skfda/ml/classification/_logistic_regression.py
```diff
@@ -0,0 +1,193 @@
...
+        X, classes, y_ind = self._argcheck_X_y(X, y)
+
+        self.classes_ = classes
+
+        n_samples = len(y)
+        n_features = len(X.grid_points[0])
+
+        selected_indexes = np.zeros(self.p, dtype=np.intc)
+
+        mvlr = mvLogisticRegression()  # multivariate logistic regression
```
vnmabus, line 94: Note that by default logistic regression in scikit-learn has a L2 regularization. If that is intended make it explicit, otherwise pass the regularization used by the original article.

vnmabus (changes requested):

skfda/ml/classification/_logistic_regression.py
```diff
@@ -0,0 +1,193 @@
...
+    for functional logistic regression proposed in
+    :footcite:ts:`bueno++_2021_functional`.
+
+    .. warning::
+        For now, only binary classification for functional
+        data with one dimensional domains is supported.
+
+    Args:
+        p: number of points (and coefficients) to be selected by
+        the algorithm.
```
vnmabus, line 31: Check the generated doc: this should be indented further.

skfda/ml/classification/_logistic_regression.py
```diff
@@ -91,7 +92,8 @@ def fit(  # noqa: D102
 
         selected_indexes = np.zeros(self.p, dtype=np.intc)
 
-        mvlr = mvLogisticRegression()  # multivariate logistic regression
+        # multivariate logistic regression
+        mvlr = mvLogisticRegression(penalty='l2')
```
vnmabus, line 96: Does the original algorithm use L2 penalization?

dSerna4: I don´t think so, but the original algorithm use firth's estimator, which we don´t use. I have tried to use no penalty but it fails (apparently there is a division by zero somewhere). Besides, in this algorithm, I don´t think it really matters which penalty we use in the multivariate part.

vnmabus (approved):
LGTM

PR #397, opened 2021-11-23
Mahalanobis

pedrorponga:


skfda/preprocessing/dim_reduction/feature_extraction/_fpca.py
```diff
@@ -370,7 +369,7 @@ def _fit_grid(
 
         basis_matrix = basis.data_matrix[..., 0]
         if regularization_matrix is not None:
-            basis_matrix = basis_matrix + regularization_matrix
+            basis_matrix += regularization_matrix
```
vnmabus, line 372: This changes the `basis` object.

vnmabus (changes requested):

vnmabus (changes requested):
Sorry for not taking a look at this before. I though that you would ping me when it was ready to review.

skfda/misc/metrics/_mahalanobis.py
```diff
@@ -0,0 +1,143 @@
...
+
+        Args:
+            e1: First object.
+            e2: Second object.
+
+        Returns:
+            Squared functional Mahalanobis distance between two observations.
+        """
+        if self.eigenvalues is not None and self.eigenvectors is not None:
+            self.eigenvalues_ = self.eigenvalues
```
vnmabus, line 133: This should be done in `fit`. If `fit` is not called, then you can use a local variable to get either `self.eigenvalues_` or `self.eigenvalues`, but you shouldn't change the object in `__call__`.

skfda/ml/classification/_centroid_classifiers.py
```diff
@@ -86,6 +86,9 @@ def fit(self, X: T, y: NDArrayInt) -> NearestCentroid[T]:
         Returns:
             self
         """
+        fit = getattr(self.metric, 'fit', lambda X: None)
+        fit(X)
```
vnmabus, line 90: If this is going to be done several times, maybe it should be in a function in utils.

vnmabus (changes requested):
BTW, the tests pass now, it was a temporary timeout error.

vnmabus (approved):

PR #399, opened 2021-11-30
Fix computation of eigenvalues in FPCA, for both basis and grid.

vnmabus:
Fixes computation of eigenvalues in FPCA.
- Added a new test comparing the results with no regularization, uniform weights and equispaced grid with the multivariate case.
- Compare singular values for grid with the ones obtained in fda.usc.
- Compare explained variance for basis with the one obtained in fda.
- Added comments with reproducibility code for the R packages.

Note: regularization is not tested in this PR.

@hzzhyj As the original author, if you have time to review the changes to ensure nothing is wrong I would be grateful.


skfda/preprocessing/dim_reduction/feature_extraction/_fpca.py
```diff
@@ -375,7 +378,7 @@ def _fit_grid(
         ).T
```
pedrorponga, line 378: Line 373: basis_matrix += regularization_matrix

vnmabus: That would change the basis.

PR #422, opened 2022-02-17
Change example of per class transformer

alvaro-castillo:


vnmabus (changes requested):
There are still issues. For example, the code:
```python
        >>> for i, curve_grid in enumerate(x_transformed2.iloc[:,0].values):
        ...     if i == 0:
        ...         X_transformed_grid = curve_grid
        ...     else:
        ...         X_transformed_grid = X_transformed_grid.concatenate(
        ...                                 curve_grid,
        ...                              )
        >>> y = np.concatenate((y,y))
```
only takes into account the first column, and it is unnecessary (as `values` is already a `FDataGrid`). It would be much most useful to concatenate both columns as coordinates in one `FDataGrid`, so that a functional distance could be directly applied taking both into account.

vnmabus (approved):

PR #425, opened 2022-02-22
Implement unary minus for FData 

### LINKED ISSUE #356: Implement unary minus for FData
The unary minus operator should be implemented for `FData` subclasses.

jiduque:
Closes #356: Implement unary minus for FData

### DIFF
--- a/skfda/representation/basis/_fdatabasis.py
+++ b/skfda/representation/basis/_fdatabasis.py
@@ -912,6 +912,11 @@ class FDataBasis(FData):  # noqa: WPS214
         """Right division for FDataBasis object."""
         return NotImplemented
 
+    def __neg__(self: T) -> T:
+        """Negation of FData object."""
+        return self._copy_op(other=None, coefficients=-self.coefficients)
+
+
     #####################################################################
     # Pandas ExtensionArray methods
     #####################################################################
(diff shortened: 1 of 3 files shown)


skfda/representation/basis/_fdatabasis.py
```diff
@@ -912,6 +912,11 @@ def __rtruediv__(
         """Right division for FDataBasis object."""
         return NotImplemented
 
+    def __neg__(self: T) -> T:
+        """Negation of FData object."""
+        return self._copy_op(other=None, coefficients=-self.coefficients)
```
vnmabus, line 917: `self._copy_op` is only really needed for binary operations. It is used to consider the case where a `FData` with `n_samples == 1` and other with `n_samples > 1` are operated together. In that case you want to copy the `FData` with `n_samples > 1`, as the result will have the same number of samples. In this case, as there is only one operand, you can directly use `self.copy` instead.

skfda/representation/grid.py
```diff
@@ -798,6 +798,10 @@ def __rtruediv__(
 
         return self._copy_op(other, data_matrix=data_matrix / self.data_matrix)
 
+    def __neg__(self: T) -> T:
+        """Negation of FData object."""
+        return self._copy_op(other=None, data_matrix=-self.data_matrix)
```
vnmabus, line 803: Same as above.

vnmabus (changes requested):
Thank you for contributing! Just a small change and it will be ready to go.

vnmabus (approved):
LGTM

PR #427, opened 2022-02-25
Paremeterized Functional QDA

alvaro-castillo:


skfda/ml/classification/_parametrized_functional_qda.py
```diff
@@ -0,0 +1,229 @@
...
+from ...representation import FDataGrid
+
+
+class ParametrizedFunctionalQDA(
+    BaseEstimator,  # type: ignore
+    ClassifierMixin,  # type: ignore
+):
+    """Parametrized functional quadratic discriminant analysis.
+
+    This classifier is based on the assumption that the data is part
```
vnmabus, line 20: I think we should add a warning in the documentation that this classifier is experimental, as it does not come from a peer-published paper. Please bring this topic to next meeting.

examples/plot_classification_methods.py
```diff
@@ -0,0 +1,132 @@
+"""
+Classification methods.
+
```
vnmabus, line 3: This empty line is wrong.

examples/plot_classification_methods.py
```diff
@@ -0,0 +1,132 @@
...
+==================================
+
+Shows a comparison between different classification methods.
```
vnmabus, line 6: Several things wrong with the example:
- Almost no explanation of the methods/why they have been chosen.
- The plotted images don't say much.
- No comparison of the accuracy of each method.
- No conclusions at all.

vnmabus (changes requested):

vnmabus (approved):

PR #438, opened 2022-04-06
Feature Construction Moments

alvaro-castillo:


skfda/exploratory/stats/_functional_transformers.py
```diff
@@ -342,3 +343,143 @@ def number_up_crossings(
...
         axis=2,
     ).T
+
+
+def unconditional_central_moments(
+    data: Union[FDataBasis, FDataGrid],
+    n: int,
+) -> NDArrayFloat:
+    """
+    Calculate the unconditional central moments of a dataset.
```
vnmabus, line 353: Give a larger explanation with the formulas, as it is not a widely known concept.

skfda/exploratory/stats/_functional_transformers.py
```diff
@@ -342,3 +343,143 @@ def number_up_crossings(
...
+    >>> import numpy as np
+    >>> from skfda.exploratory.stats import unconditional_central_moments
+    >>> np.around(unconditional_central_moments(X[:5], 1), decimals=2)
+        array([[ 0.02, -0.02],
+               [ 0.03, -0.02],
+               [ 0.03, -0.01],
+               [ 0.02, -0.  ],
+               [ 0.03,  0.01]])
+    """
+    mean = np.mean(data.data_matrix, axis=1)
```
vnmabus, line 384: This don't work for `FDataBasis`.

skfda/exploratory/stats/_functional_transformers.py
```diff
@@ -342,3 +343,143 @@ def number_up_crossings(
...
+               [ 4.88],
+               [ 4.85],
+               [ 4.9 ],
+               [ 4.84]])
+    """
+    domain_range = data.domain_range[0][1] - data.domain_range[0][0]
+
+    if isinstance(data, FDataGrid):
+        return scipy.integrate.simpson(
+            function(data.data_matrix),
```
vnmabus, line 476: You can apply the function directly to the `FDataGrid` and use its `integrate` method afterwards.

skfda/exploratory/stats/_functional_transformers.py
```diff
@@ -342,3 +343,143 @@ def number_up_crossings(
...
+    Then we call the function with the dataset and the function.
+    >>> from skfda.exploratory.stats import unconditional_expected_value
+    >>> np.around(unconditional_expected_value(X[:5], f), decimals=2)
+        array([[ 4.96],
+               [ 4.88],
+               [ 4.85],
+               [ 4.9 ],
+               [ 4.84]])
+    """
+    domain_range = data.domain_range[0][1] - data.domain_range[0][0]
```
vnmabus, line 472: Should this be extended to general functions?

skfda/exploratory/stats/_functional_transformers.py
vnmabus: Also, it should be the mean in the integral sense, IMHO, to consider the position of the grid points too.

vnmabus (changes requested):

skfda/exploratory/stats/_functional_transformers.py
alvaro-castillo: Then the function should be a Callable[[FDataGrid], FDataGrid] so a FDataGrid is returned and the integrate method could be applied later

vnmabus: You can apply a `ndarray` function over a `FDataGrid`. Mypy probably won't like it but it will work at runtime.

skfda/exploratory/stats/_functional_transformers.py
```diff
@@ -342,3 +342,154 @@ def number_up_crossings(
...
+
+def unconditional_moments(
+    data: Union[FDataBasis, FDataGrid],
+    n: int,
+) -> NDArrayFloat:
+    r"""
+    Calculate the specified unconditional moment of a dataset.
+
+    It performs the following map:
+    :math:`f(X)=\int_a^b f_1(X(t))dt,\dots,f_p(X(t))dt=\int_a^b  f^p(X(t))dt`.
```
vnmabus, line 409: Where is the `n` in this formula? What is `p`?

vnmabus (changes requested):

skfda/exploratory/stats/_functional_transformers.py
```diff
@@ -342,3 +342,167 @@ def number_up_crossings(
...
+        [
+            (iterval[1] - iterval[0])
+            for iterval in data.domain_range
+        ],
+    )
+
+    if isinstance(data, FDataGrid):
+        return function(data).integrate() / domain_range
+
+    return nquad_vec(
```
vnmabus, line 505: I don't think this is ok. You seem to compute the integral of the function instead of the function applied to the random variable. Please add a test using the same data in both discretized and basis form to check if it works.

vnmabus (changes requested):

PR #440, opened 2022-04-18
Feature construction Function Transformers

alvaro-castillo:


skfda/preprocessing/_function_transformers.py
```diff
@@ -0,0 +1,250 @@
...
+                [ 163.53],
+                [ 162.29],
+                [ 153.1 ],
+                [ 178.48],
+                [ 161.75]]])
+
+    """
+
+    def __init__(self, n_intervals: int):
+        self.n_intervals = n_intervals
```
vnmabus, line 102: Instead of adding parameters and overloading the `transform` function, just call the parent `__init__` with the appropriate arguments.

The alternative is not to inherit from `FunctionTransformer` and just make this a fully-fledged transformer, but that looks like an overkill to me.

vnmabus (changes requested):

PR #442, opened 2022-04-24
Function transformers

alvaro-castillo:


skfda/preprocessing/feature_construction/_function_transformers.py
```diff
@@ -0,0 +1,196 @@
...
+
+        Then we decide how many intervals we want to consider (in our case 2)
+        and call the function with the dataset.
+        >>> import numpy as np
+        >>> from skfda.preprocessing.feature_construction import (
+        ...     LocalAveragesTransformer,
+        ... )
+        >>> local_averages = LocalAveragesTransformer(2)
+        >>> np.around(local_averages.fit_transform(X), decimals=2)
+        array([[[ 116.94],
```
vnmabus, line 40: This should be a 2D matrix to be able to use it in further scikit-learn estimators.

skfda/preprocessing/feature_construction/_function_transformers.py
```diff
@@ -0,0 +1,196 @@
...
+    >>> from scipy.special import jv
+    >>> from skfda.representation import FDataGrid
+    >>> import numpy as np
+    >>> x_grid = np.linspace(0, 14, 14)
+    >>> fd_grid = FDataGrid(
+    ...     data_matrix=[jv([0], x_grid)],
+    ...     grid_points=x_grid,
+    ... )
+    >>> fd_grid.data_matrix
+    array([[[ 1.        ],
```
vnmabus, line 161: This should also need to be a 2D matrix.

vnmabus (changes requested):

skfda/preprocessing/feature_construction/_function_transformers.py
alvaro-castillo: The output of OccupationMeasureTransformer is indeed a 2D array. What you are pointing is the data_matrix of a Bessel Function on a FDataGrid format.

vnmabus: Sure thing! Sorry for the mistake.

skfda/exploratory/stats/_functional_transformers.py
```diff
@@ -127,7 +67,7 @@ def local_averages(
         data.integrate(interval=((intervals[i], intervals[i + 1]))) / step
         for i in range(n_intervals)
     ]
-    return np.asarray(integrated_data)
+    return np.asarray(integrated_data)[:, :, 0]
```
vnmabus, line 70: I am actually fine with this function returning a ND-array. It is the transformer, the one which has to return a 2D array. And I think that has to be done using `reshape` and not getting only the 0-th element, as that won't work for more general functions.

skfda/exploratory/stats/_functional_transformers.py
```diff
@@ -34,85 +34,25 @@ def local_averages(
         calculate the local averages.
         n_intervals: number of intervals we want to consider.
     Returns:
-        ndarray of shape (n_intervals, n_samples, n_dimensions)
-        with the transformed data for FDataBasis and (n_intervals, n_samples)
-        for FDataGrid.
+        ndarray of shape (n_intervals, n_samples) with the
```
vnmabus, line 37: I think that the first dimension should always be the number of samples, for consistency and compatibility with scikit-learn tools.

skfda/preprocessing/feature_construction/_function_transformers.py
```diff
@@ -37,12 +37,8 @@ class LocalAveragesTransformer(BaseEstimator, TransformerMixin):
         ... )
         >>> local_averages = LocalAveragesTransformer(2)
         >>> np.around(local_averages.fit_transform(X), decimals=2)
-        array([[[ 116.94],
-                [ 111.86],
-                [ 107.29]],
-               [[ 177.26],
-                [ 157.62],
-                [ 154.97]]])
+        array([[ 116.94,  111.86,  107.29],
```
vnmabus, line 40: The number of rows should always be the number of samples.

vnmabus (changes requested):

skfda/preprocessing/feature_construction/_function_transformers.py
alvaro-castillo: `occupation_measure` then should also be changed to comply with shape(n_samples, n_intervals)

skfda/exploratory/stats/_functional_transformers.py
```diff
@@ -127,7 +71,7 @@ def local_averages(
         data.integrate(interval=((intervals[i], intervals[i + 1]))) / step
         for i in range(n_intervals)
     ]
-    return np.asarray(integrated_data)
+    return np.transpose(integrated_data, (1, 0, 2))
 
 
 def _calculate_curves_occupation_(
```
vnmabus, line 77: Why the last underscore?

skfda/exploratory/stats/_functional_transformers.py
```diff
@@ -254,7 +199,7 @@ def occupation_measure(
         function_y_coordinates,
         function_x_coordinates,
         intervals,
-    )
+    ).T
```
vnmabus, line 202: Couldn't the computation be done already using the right dimensions?

vnmabus (changes requested):

PR #449, opened 2022-05-17
Add rst

alvaro-castillo:


docs/modules/preprocessing.rst
```diff
@@ -39,4 +39,14 @@ the data with clarity. To better understand the data, we need to use
...
 while still preserving the most relevant information.
 :doc:`Here <preprocessing/dim_reduction>` you can learn more about the
-dimension reduction methods available in the library.
\ No newline at end of file
+dimension reduction methods available in the library.
+
+Feature construction
+--------------------
+
+It is crucial to create features from the original ones so they can be
```
vnmabus, line 47: "crucial" is maybe too exaggerated. It is an important technique, for sure, but not "crucial".

docs/modules/preprocessing/feature_construction.rst
```diff
@@ -0,0 +1,32 @@
...
+   :toctree: autosummary
+
+   skfda.preprocessing.feature_construction.FDAFeatureUnion
+
+
+Per class transformer
+---------------------
+
+This method deals with the extraction of features depending on the class
+label of each observation. It will apply a different transformation to each
```
vnmabus, line 26: That is a very misleading description. It applies as many transformation as classes to EVERY observation. Each transformation is fitted using only the training data of a particular class.

vnmabus (changes requested):

vnmabus (approved):

PR #452, opened 2022-06-07
Adding a comparison table to the example

alvaro-castillo:


examples/plot_classification_methods.py
```diff
@@ -138,6 +138,19 @@
 # achieve more than an 80% of score, but the most robust ones are KNN and
 # Parameterized Functional QDA.
 
+##############################################################################
```
vnmabus, line 141: Can't this be generated programatically? I don't want it to be desynchronized with the code.

vnmabus (changes requested):

vnmabus (approved):

PR #458, opened 2022-06-22
Handle ops between FDGrid and Python functions

matthieubulte:
Hi there 👋

The goal of this PR is to be able to make it easier to do operations between FDGrids and normal python functions, like this
```python
fd + (lambda x: x**2)
```

The feature seems to work and tests are still green but the PR probably needs a little bit more work to be mergeable (I don't think I manage well all possible grids since it's not clear to me what is possible to have as a grid), so I would be happy to get guidance on this.

Cheers!

### DIFF
--- a/skfda/representation/grid.py
+++ b/skfda/representation/grid.py
@@ -711,7 +711,7 @@ class FDataGrid(FData):  # noqa: WPS214
 
     def _get_op_matrix(
         self,
-        other: Union[T, NDArrayFloat, NDArrayInt, float, Callable],
+        other: Union[T, NDArrayFloat, NDArrayInt, float, Callable[[Union[float, NDArrayFloat]], Union[float, NDArrayFloat]]],
     ) -> Union[None, float, NDArrayFloat, NDArrayInt]:
         if isinstance(other, numbers.Real):
             return float(other)
@@ -742,13 +742,14 @@ class FDataGrid(FData):  # noqa: WPS214
             return other.data_matrix
 
         elif isinstance(other, Callable):
-            return np.array([[other(x) for x in _gp] for _gp in self.grid_points]).T
+            coordinates = np.array(np.meshgrid(*self.grid_points, indexing='ij')).T.reshape(-1, self.dim_domain)
+            return np.array([other(x) for x in coordinates]).reshape((1,) + self.data_matrix.shape[1:])
 
         return None
 
     def __add__(
         self: T,
-        other: Union[T, NDArrayFloat, NDArrayInt, float, Callable],
+        other: Union[T, NDArrayFloat, NDArrayInt, float, Callable[[Union[float, NDArrayFloat]], Union[float, NDArrayFloat]]],
     ) -> T:
 
         data_matrix = self._get_op_matrix(other)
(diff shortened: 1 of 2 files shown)


skfda/representation/grid.py
```diff
@@ -710,12 +711,11 @@ def __eq__(self, other: object) -> NDArrayBool:  # type: ignore[override]
 
     def _get_op_matrix(
         self,
-        other: Union[T, NDArrayFloat, NDArrayInt, float],
+        other: Union[T, NDArrayFloat, NDArrayInt, float, Callable],
```
vnmabus, line 714: We shouldn't accept arbitrary callables. Please include the expected input and output types.

skfda/representation/grid.py
```diff
@@ -741,11 +741,14 @@ def _get_op_matrix(
             self._check_same_dimensions(other)
             return other.data_matrix
 
+        elif isinstance(other, Callable):
+            return np.array([[other(x) for x in _gp] for _gp in self.grid_points]).T
```
vnmabus, line 745: A couple of things:
- I think we should limit support only to vectorized functions. Otherwise we have to call repeteadly the function for every point, which is too slow.
- I think this code won't work for functions of several variables (e.g. surfaces). For supporting that case we should probably create a matrix with every coordinate were functions are going to be evaluated, call the function with that matrix and then reshape the result.
- Another option would be to check first if the function supports vectorized input, and fall back to the slow point by point evaluation otherwise. This would be (much) less performant, but probably useful for newcomers and small datasets.
- `FData` are callables too. What should happen if we do ops between `FData` with different domains? And between two `FDataGrid` with the same domain and different `grid_points`? (we could lose commutativity in this case!) And between a `FDataGrid` and a `FDataBasis` with the same domain?

tests/test_grid.py
```diff
@@ -161,6 +160,10 @@ def test_add(self):
         np.testing.assert_array_equal(fd2.data_matrix[..., 0],
                                       [[3, 4, 5, 6], [4, 5, 6, 7]])
 
+        fd2 = fd1 + (lambda x: x)
```
vnmabus, line 163: We would need to test in addition:
- Behaviour for functions of several variables.
- Behaviour for vector-valued functions.
- A combination of all the above.

vnmabus (changes requested):

skfda/representation/grid.py
matthieubulte: 1. My understanding is that it doesn't make a real difference in terms of performance except for very small arrays, see [1].
2/3. This is true, it's now fixed.
4. This is also true, and a good point, which is why the testing for `Callable` is after testing for `FDataGrid`. That way all your concerns are tackled by the `FDataGrid` handler.

[1] https://stackoverflow.com/questions/35215161/most-efficient-way-to-map-function-over-numpy-array

matthieubulte:
I have added a commit to address most of your comments, let me know what you think!

skfda/representation/grid.py
vnmabus: Sorry for the delay.

> 1. My understanding is that it doesn't make a real difference in terms of performance except for very small arrays, see [1].

I didn't mean to use `np.vectorize`, but to assume that the `other` function accepts an array as input and return the array as output. This way, it would be very efficient for ufuncs like `np.sqrt` or `np.sin`, or functions created composing these, as they perform the iteration in C code, probably using SIMD instructions for further performance.

>        2/3. This is true, it's now fixed.

I would first try passing the  array of coordinates, and if that fails pass the coordinates as different parameters for the efficiency reasons explained above.

>     2. This is also true, and a good point, which is why the testing for `Callable` is after testing for `FDataGrid`. That way all your concerns are tackled by the `FDataGrid` handler.

Ok, so for now `FDataGrid` is special, and two `FDataGrid`s with different domains cannot be operated together. If a user wants that they can always use `lambda x: my_fdatagrid(x)` instead. We can also change this behaviour in the future if we found a reasonable way to support it.

PR #466, opened 2022-08-07
Feature/fpca_regression

Ddelval:
Add the FPCA regression estimator


vnmabus (changes requested):
I think this functionality is implemented in the package `fda.usc`. We should have a test verifying that we obtain the same results.

Ddelval:
> I think this functionality is implemented in the package `fda.usc`. We should have a test verifying that we obtain the same results.

I have (finally) added this test. However, while going back and forth between python and R, I stumbled upon a python packet that provides a simple way to execute R code from within python. I am sceptical regarding whether or not this could be useful for the tests. On the one hand, it would ensure that the tests are testing against the correct values while forcing us to include the R source used to obtain the test reference data. On the other hand, it would make the tests considerably slower and require modifications to the GitHub actions.

I think the timing tradeoff is too significant (I measured a slowdown of around 6x). Nevertheless, I thought I might just mention it just in case it comes in handy at some point. The following snippet generates the testing data for the test I just added.
```
from rpy2 import robjects

Rresult = robjects.r('''
library("fda.usc")
data(tecator)
# Fit the regression model with the first 129 observations
x=tecator$absorp.fdata[1:129,]
y=tecator$y$Fat[1:129]
res2=fregre.pc(x,y,l=1:10)

# Predict the response for the remaining observations
n = length(tecator$y$Fat)
xnew=tecator$absorp.fdata[130:n,]
result = predict(res2, xnew)
names(result) = NULL

# Output the predicted values
paste(
    round(result,8),
    collapse = ", "
)
''')
r_predictions = np.array(Rresult[0].split(", "), dtype=np.float64)
```

skfda/ml/regression/_fpca_regression.py
```diff
@@ -0,0 +1,150 @@
...
+
+        X_transformed = FDataBasis(
+            basis=self.fpca_basis,
+            coefficients=transformed_coefficients,
+        )
+        self._linear_model.fit(X_transformed, y)
+
+        self.n_components_ = self.n_components
+        self.components_ = self._fpca.components_
+        self.coefficients_ = self._linear_model.coef_
```
vnmabus, line 124: Why don't you use the same name?

skfda/ml/regression/_fpca_regression.py
```diff
@@ -0,0 +1,150 @@
...
+            regression. If None then no regularization is applied.
+            Defaults to ``None``.
+        components_basis: Basis used for the principal components. If None
+            then the basis of the input data is used. Defaults to None.
+            It is only used if the input data is a FDataBasis object.
+
+    Attributes:
+        n\_components\_: Number of principal components used.
+        components\_: Principal components.
+        cofficents\_: Coefficients of the linear regression model.
```
vnmabus, line 45: There is a typo here.
```suggestion
        coefficients\_: Coefficients of the linear regression model.
```

skfda/ml/regression/_fpca_regression.py
```diff
@@ -0,0 +1,150 @@
...
+        >>> score = reg.score(fd, y)
+        >>> reg.predict(fd) # doctest:+ELLIPSIS
+        array([...])
+
+    """
+
+    def __init__(
+        self,
+        n_components: int = 2,
+        intercept: bool = True,
```
vnmabus, line 72: Why not `fit_intercept`?

skfda/ml/regression/_fpca_regression.py
```diff
@@ -0,0 +1,150 @@
...
+        self.n_components = n_components
+        self.intercept = intercept
+        self.pca_regularization = pca_regularization
+        self.regression_regularization = regression_regularization
+        self.components_basis = components_basis
+
+    def fit(
+        self,
+        X: FData,
+        y: np.ndarray,
```
vnmabus, line 86: Use a more specific type if possible.

skfda/ml/regression/_fpca_regression.py
```diff
@@ -0,0 +1,150 @@
...
+        self.coefficients_ = self._linear_model.coef_
+        self.explained_variance_ = self._fpca.explained_variance_
+        self.explained_variance_ratio_ = self._fpca.explained_variance_ratio_
+
+        return self
+
+    def predict(
+        self,
+        X: FData,
+    ) -> np.ndarray:
```
vnmabus, line 133: Use a more specific type.

skfda/representation/basis/_custom_basis.py
```diff
@@ -195,6 +195,22 @@ def _evaluate(
...
+        Subclasses may override this method for improving computation
+        of the Gram matrix.
+
+        """
+        if isinstance(self.fdata, FDataBasis):
+            basis_gram = self.fdata.basis.gram_matrix()
+            coefficients = self.fdata.coefficients
+            return coefficients @ basis_gram @ coefficients.T
+
+        from ...misc import inner_product_matrix
```
vnmabus, line 211: Do the import at the beginning of the function (just after the docstring).

skfda/tests/test_fpca_regression.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+class FPCARegressionTestCase(unittest.TestCase):
+    """Tests for principal component analysis."""
+
+    def test_fpca_reg_against_fda_usc(self) -> None:
+        """Check that the results obtained are similar to those of fda.usc.
+
+        Results obtained from fda.usc with the following R code
+            library("fda.usc")
```
vnmabus, line 23: Were you able to compare results with refund too?

vnmabus (changes requested):
Remember to also add the class to the API reference documentation.

skfda/tests/test_fpca_regression.py
Ddelval: This is the issue that I opened: https://github.com/refunders/refund/issues/102.
It doesn't seem like we'll be getting any solutions from their side anytime soon.

skfda/representation/basis/_custom_basis.py
```diff
@@ -195,6 +195,23 @@ def _evaluate(
...
+        """
+        Compute the Gram matrix.
+
+        Subclasses may override this method for improving computation
+        of the Gram matrix.
+
+        """
+        from ...misc import inner_product_matrix
+
+        if isinstance(self.fdata, FDataBasis):
```
vnmabus, line 208: Why is this specialized for `FDataBasis`?

skfda/ml/regression/_fpca_regression.py
```diff
@@ -0,0 +1,152 @@
...
+        """Predict using the linear model.
+
+        Args:
+            X: Functional data.
+
+        Returns:
+            Target values.
+
+        """
+        check_is_fitted(self, ["_fpca", "_linear_model"])
```
vnmabus, line 145: I think you do not need to pass the second argument to this function.

skfda/representation/basis/_custom_basis.py
Ddelval: Leftover from the previous implementation. Fixed now.

vnmabus (approved):

PR #467, opened 2022-08-08
Documentation fixes, weights deprecation and linter warnings

Ddelval:
The following improvements have been made:
- Improve the documentation of `centering`, `components_` and `weights`
- Mark the `weights` as deprecated
- Ensure that the code follows pep-8. In particular, starting non-public fields with an underscore


skfda/preprocessing/dim_reduction/_fpca.py
```diff
@@ -99,11 +101,16 @@ def __init__(
         weights: Optional[Union[ArrayLike, WeightsCallable]] = None,
         components_basis: Optional[Basis] = None,
     ) -> None:
-        self.n_components = n_components
-        self.centering = centering
-        self.regularization = regularization
-        self.weights = weights
-        self.components_basis = components_basis
+        self._n_components = n_components
```
Ddelval, line 104: @vnmabus, I changed the names of the "private" fields so that they start with an underscore since the linter counted them as public and complained about the number of public fields.
However, it might be a good idea to make the number of componentes an attribute. I understand it can be obtained from `componets_` but it seems to me that it is the sort of value that should be easily available

skfda/preprocessing/dim_reduction/_fpca.py
```diff
@@ -37,23 +36,26 @@ class FPCA(
     Parameters:
         n_components: Number of principal components to keep from
             functional principal component analysis. Defaults to 3.
-        centering: If ``True`` then calculate the mean of the functional
-            data object and center the data first. Defaults to ``True``.
+        centering: Set to ``False`` when the functional data is already known
+            to be centered and there is no need to center it. Otherwise,
+            the mean of the functional data object is calculated and, the data,
```
vnmabus, line 41:
```suggestion
            the mean of the functional data object is calculated and the data
```

skfda/preprocessing/dim_reduction/_fpca.py
```diff
@@ -37,23 +36,26 @@ class FPCA(
...
             weight vector will be obtained by evaluating the object at the
             sample points of the passed FDataGrid object in the fit method.
             This parameter is only used when fitting a FDataGrid.
 
+            .. deprecated:: 0.7.2
+
     Attributes:
-        components\_ (FData): this contains the principal components in a
-            basis representation.
+        components\_ (FData): this contains the principal components.
```
vnmabus, line 58: Remove the types from the docstring.

skfda/preprocessing/dim_reduction/_fpca.py
vnmabus: These fields are not private. Also, a `scikit-learn`-compatible estimator should attempt to do the bare minimum in `__init__` (just store the parameters).

skfda/preprocessing/dim_reduction/_fpca.py
```diff
@@ -99,11 +101,30 @@ def __init__(
...
+
+        """
+        self._n_components = n_components
+        self._centering = centering
+        self._regularization = regularization
+        self._weights = weights
+        self._components_basis = components_basis
+        if weights is not None:
+            warnings.warn(
+                "The weights parameter is deprecated and will be removed.",
```
vnmabus, line 125:
```suggestion
                "The 'weights' parameter is deprecated and will be removed.",
```

vnmabus (changes requested):

skfda/preprocessing/dim_reduction/_fpca.py
Ddelval: Okay @vnmabus . I thought the convention was to label public attributes with a trailing underscore. Now I am not sure why some of the fields have that trailing underscore, and others do not. 

If I revert the changes, I will get the `WPS230: Found too many public instance attributes: 11 > 10` warning again
- I can add a noqa comment and ignore it
- I could increase the number of maximum public instance attributes to 11 in the `setup.cfg`

Which one do you think is a better option?

vnmabus: The trailing underscore is for public attributes set during `fit` (see https://scikit-learn.org/stable/developers/develop.html).

I think # noqa (with the appropriate error code).

Ddelval: Oh, that makes sense.

Great. I will add it in my next push.

Thanks!

skfda/preprocessing/dim_reduction/_fpca.py
```diff
@@ -99,11 +101,30 @@ def __init__(
         weights: Optional[Union[ArrayLike, WeightsCallable]] = None,
         components_basis: Optional[Basis] = None,
     ) -> None:
+        """
```
vnmabus, line 104: Don't document `__init__`, we already document it in the class.

vnmabus (changes requested):

vnmabus (approved):
LGTM

PR #481, opened 2022-10-06
Issue/classifier classes/test

m5signorini:
Tests for securing the existence of the `classes_` attribute in a classifier after calling the `fit` method.

### DIFF
--- a/skfda/tests/test_fpca.py
+++ b/skfda/tests/test_fpca.py
@@ -412,6 +412,13 @@ class FPCATestCase(unittest.TestCase):
 
         fpca = FPCA(n_components=n_components, _weights=[1] * 365)
         fpca.fit(fd_data)
+
+        # The fda.usc uses the trapezoid rule to compute the integral
+        # with the following weights
+        weights = np.ones(len(fd_data.grid_points[0]))
+        weights[0] = 0.5
+        weights[-1] = 0.5
+        fpca._weights = weights  # noqa: WPS437 (protected access)
         scores = fpca.transform(fd_data)
 
         # results obtained
@@ -427,7 +434,7 @@ class FPCATestCase(unittest.TestCase):
             216.43002289, 233.53770292, 344.18479151,
         ])
 
-        np.testing.assert_allclose(scores.ravel(), results, rtol=0.25)
+        np.testing.assert_allclose(scores.ravel(), results)
 
     def test_grid_fpca_regularization_fit_result(self) -> None:
         """Compare the components in grid against the fda.usc package."""


skfda/tests/test_classifier_classes.py
```diff
@@ -0,0 +1,76 @@
...
+        """Establish train and test data sets."""
+        X, y = fetch_growth(return_X_y=True)
+        X_train, X_test, y_train, _ = train_test_split(
+            X,
+            y,
+            test_size=0.25,
+            stratify=y,
+            random_state=0,
+        )
+        self._X_train = X_train
```
vnmabus, line 44: As test classes are not to be used by users, you can made their fields public without problem. That way it is more legible.

skfda/tests/test_classifier_classes.py
```diff
@@ -0,0 +1,76 @@
...
+                    Gaussian(variance=6, length_scale=1),
+                ),
+            ),
+        ]
+
+    def test_classes(self) -> None:
+        """Check classes attribute."""
+        # Iterate over all classifiers with index
+        for i, clf in enumerate(self.tested_classifiers):
+            with self.subTest(i=i):
```
vnmabus, line 70: The values passed in `subTest` are used for displaying purposes if the test fails. The number of the iteration is not very informative in that regard. Use the class name instead.

skfda/tests/test_classifier_classes.py
```diff
@@ -0,0 +1,76 @@
...
+            test_size=0.25,
+            stratify=y,
+            random_state=0,
+        )
+        self._X_train = X_train
+        self._X_test = X_test
+        self._y_train = y_train
+
+        self.classes = _classifier_get_classes(self._y_train)[0]
+        self.tested_classifiers: list[ClassifierMixin[FData, NDArrayAny]] = [
```
vnmabus, line 49: Until we cease support for Python 3.8 you need to use `List` instead (imported from `typing`).

vnmabus, line 49: Right now, the tests won't pass as no fixes have been done, so I cannot accept it. I propose to leave the list empty and then adding each fixed classifier in their corresponding PR.

skfda/tests/test_classifier_classes.py
```diff
@@ -0,0 +1,76 @@
...
+from ..typing._numpy import NDArrayAny
+
+
+class TestClassifierClasses(unittest.TestCase):
+    """Test for classifiers classes."""
+
+    def setUp(self) -> None:
+        """Establish train and test data sets."""
+        X, y = fetch_growth(return_X_y=True)
+        X_train, X_test, y_train, _ = train_test_split(
```
vnmabus, line 37: For this test we don't actually require splitting, do we?

skfda/tests/test_classifier_classes.py
```diff
@@ -0,0 +1,76 @@
...
+
+from ..typing._numpy import NDArrayAny
+
+
+class TestClassifierClasses(unittest.TestCase):
+    """Test for classifiers classes."""
+
+    def setUp(self) -> None:
+        """Establish train and test data sets."""
+        X, y = fetch_growth(return_X_y=True)
```
vnmabus, line 36: I would use synthetic data for this test, as the actual functions are not used.
I would also try a case where `y` consists of string labels and another were numbers are used instead.

vnmabus (changes requested):

skfda/tests/test_classifier_classes.py
```diff
@@ -0,0 +1,59 @@
...
+        """Check classes attribute."""
+        # Iterate over all classifiers with index
+        for clf in self.tested_classifiers:
+            # Iterate over all class sets to test different types of classes
+            for classes, y in zip(self.classes_list, self.y_list):
+                message = (
+                    f'Testing classifier {clf.__class__.__name__} '
+                    f'with classes {classes}'
+                )
+                with self.subTest(msg=message):
```
vnmabus, line 53: This is enough:
```suggestion
                with self.subTest(classifier=clf, classes=classes):
```

skfda/tests/test_classifier_classes.py
```diff
@@ -0,0 +1,59 @@
...
+            np.array(["class_a", "class_b"]),
+        ]
+
+        # Create one target y data of length n_samples from each class set
+        n_samples = 30
+        self.y_list = [
+            np.resize(classes, n_samples)
+            for classes in self.classes_list
+        ]
+        self.X = make_multimodal_samples(
```
vnmabus, line 34: I would rather use `make_gaussian process` unless you really need something that only `make_multimodal_samples` could provide. Gaussian processes are the equivalent to the normal distribution for functional data. The multimodal samples are used only in registration examples.

vnmabus (changes requested):

vnmabus (approved):

PR #485, opened 2022-10-12
Fix FPCA grid implementation with regularization

Ddelval:
It has not been fixed completely, as there seems to be an issue with the penalization factor. However, now the components obtained are not orthogonal.

This was achieved by performing Cholesky decomposition. In doing this Cholesky decomposition, we can change the data to a new basis before fitting it and undo the change after calling the multivariate PCA.


vnmabus:
Is this still a draft?

Ddelval:
@vnmabus, I just forgot to mark it as ready for review.

vnmabus (approved):
LGTM

Issue #480, opened 2022-10-04
Not all classifiers have a `classes_` attribute


vnmabus:
**Describe the bug**
According to the scikit-learn [classifier glossary entry](https://scikit-learn.org/stable/glossary.html#term-classifier), every classifier should accept non-numerical classes, such as labels, and have a `classes_` attribute after `fit`, with the mapping from class names to attributes. As we want to imitate their API, this should be done also for our classifiers.

Currently we have an internal utility function `_classifier_get_classes` that uses [`LabelEncoder`](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.LabelEncoder.html#sklearn.preprocessing.LabelEncoder) to get that mapping, but in some cases it is not used and in others the attribute is not exposed with that name.

The type hints should also be fixed to take that into account. Ideally, we should attempt to leverage the type system to give an error if we forget to call `_classifier_get_classes`, for future classifiers.

We should check each classifier and fix them in separate PRs when possible:

- [x] KNeighborsClassifier
- [x] RadiusNeighborsClassifier
- [x] NearestCentroid
- [x] DTMClassifier
- [x] MaximumDepthClassifier
- [x] DDClassifier
- [x] DDGClassifier
- [x] LogisticRegression
- [x] QuadraticDiscriminantAnalysis

We should check that this behaviour is correct in each classifier. For that purpose, a test should be made that can then be called for each estimator. This can be done with either unittest or pytest, whatever is easier.

- [x] Add test

vnmabus:
It seems that this is already fixed.

PR #491, opened 2022-11-05
Issue/classifier_classes/kneighbours

m5signorini:
Solution of issue #480 for `KNeighborsClassifier` and `RadiusNeighborsClassifier`.

The `fit` method from their parent classifier class has been updated to include the initialization of the `classes_`attribute. To be able to use the corresponding functionality from `_utils`, the Target generic has been modified to a TargetClassification type var bounded to NDArrayInt and NDArrayStr, to be able to work with both string and integer class types.

### DIFF
--- a/skfda/ml/_neighbors_base.py
+++ b/skfda/ml/_neighbors_base.py
@@ -17,21 +17,29 @@ from .._utils._sklearn_adapter import (
     ClassifierMixin,
     RegressorMixin,
 )
+from .._utils._utils import _classifier_get_classes
 from ..misc.metrics import l2_distance
 from ..misc.metrics._utils import _fit_metric
-from ..representation import FData, FDataGrid, concatenate
+from ..representation import FData, concatenate
 from ..typing._metric import Metric
-from ..typing._numpy import NDArrayFloat, NDArrayInt
+from ..typing._numpy import NDArrayFloat, NDArrayInt, NDArrayStr
 
 FDataType = TypeVar("FDataType", bound="FData")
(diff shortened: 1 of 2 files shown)


skfda/ml/_neighbors_base.py
```diff
@@ -537,6 +545,31 @@ def predict_proba(
...
+                the multivariate response case.
+
+        Returns:
+            Self.
+
+        Note:
+            This method adds the attribute `classes_` to the classifier.
+        """
+        self.classes_ = _classifier_get_classes(y)[0]
+        return super().fit(X, y)
```
vnmabus, line 571: Shouldn't the passed `y` be the parsed one?

m5signorini: In the KNeighbours classes there really is no encoding and the predicted values (after `predict`), have the correct labels in `classes_`.
However, this is not the case for logistic regression, because the underlying `fit` method receives `y_ind`, so the predicted values are the encoded index of the labels. 

See
https://github.com/GAA-UAM/scikit-fda/blob/566e3e3f8e87e4ccf51cf71c538da57feefe4d74/skfda/ml/classification/_logistic_regression.py#L108-L182
where the resulting `self._mlvr` is the wrapped predictor, but it gets fitted with the indices, as shown in L174, instead of the actual classes.

I propose to solve this problem of logistic regression in another pull request. I will modify the test script to exclude logistic regression for now, and include the assertion between `classes_` and the unique values of `predict(X)`.

vnmabus (approved):

PR #496, opened 2022-11-18
Issue/classifier_classes/logreg

m5signorini:
Solution for logistic regression of issue #480.
I think it was only needed to change the last `mlvr.fit` in the fit method of logistic regression.

I also include here the corresponding updated `test_classifier_classes`, to test for all now correctly working classifiers.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -114,7 +114,7 @@ rst-directives =
     versionadded,versionchanged,
 
 rst-roles =
-    attr,class,doc,footcite,footcite:ts,func,meth,mod,obj,ref,term,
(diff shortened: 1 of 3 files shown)


m5signorini:
I have not modified the class docstring but it appears to be failing the style test. Presumably it is because the use of `:external:class` to reference scikit learn classes. Should it be instead something like only `:class:`?

vnmabus (dismissed):

vnmabus (approved):

m5signorini:
I have updated the corresponding files to avoid the style errors.
Also, I include in this pull request the changes to the `Target` type var in `_utils`.

Issue #470, opened 2022-08-26
ANOVA slow for FDataBasis


vnmabus:
ANOVA is slow for FDataBasis causing unnecessary slowdowns in tests, due to the fine grid used for GP generation. Consider generating GPs directly in basis form if possible, to achieve faster and more accurate results.

PR #505, opened 2023-01-10
Covariance function as tensor product

m5signorini:
Replacing the `cov` method of `FDataBasis` to use tensor basis instead of a numerical approximation.
This method should calculate the unbiased covariance function, which, given a list of univariate data $\\{x_n\\}_{n=1}^N$, it is defined as

$$K(t,s) =\frac{1}{N-1} \sum_{n=1}^N (x_n(t) - \bar{x}(t))(x_n(s) - \bar{x}(s)).$$

If the functions $x_n$ are expressed in a basis $\\{\phi\\}^J_{j=1}$ such that $x_n(t) = \sum \alpha_j^{(n)}\phi_j(t)$ then

$$
    K (t,s) = \frac{1}{N-1} \sum_{n=1}^N
    \left( \sum_{j=1}^J(\alpha_j^{(n)} - \bar{\alpha}_j)\phi_j(t)\right)
    \left( \sum_{j=1}^J(\alpha_j^{(n)} - \bar{\alpha}_j)\phi_j(s)\right),
$$

which it translates into calculating the covariance matrix of the coefficients $\alpha$ centered. Thus,

$$
K(t,s) = \Phi(t)^T C \Phi(s),
$$

where $\Phi$ is the basis functions as a column vector and $C$ is the unbiased sample covariance matrix of the coefficients. Thus $K$ has the matrix $C$ as coefficients in the tensor basis $\\{\phi_1(t)\phi_1(s),\quad \phi_1(t)\phi_2(s), \quad \cdots, \quad \phi_J(t)\phi_J(s)\\}$


m5signorini:
Modifying the `cov` affects tests that were assuming an `FDataGrid` as the result from calling it.
I believe this can be solved converting the result from `cov` using `to_grid` in such cases.

skfda/inference/anova/_anova_oneway.py
```diff
@@ -209,11 +209,14 @@ def _anova_bootstrap(
     random_state = validate_random_state(random_state)
 
     if equal_var:
-        k_est = concatenate(fd_grouped).cov().data_matrix[0, ..., 0]
+        k_est = concatenate(fd_grouped).cov().to_grid().data_matrix[0, ..., 0]
```
vnmabus, line 212: Related with  #470. We need to check if it is possible to generate Gaussian data with the same covariance of a given `FDataBasis` by using the coefficients directly, instead of this costly approach.

vnmabus (changes requested):
We would also need test for this behaviour.

skfda/representation/basis/_fdatabasis.py
```diff
@@ -461,25 +461,38 @@ def var(self: T, eval_points: Optional[NDArrayFloat] = None) -> T:
         """
         return self.to_grid(eval_points).var().to_basis(self.basis)
 
-    def cov(self, eval_points: Optional[NDArrayFloat] = None) -> FData:
+    def cov(self: T) -> T:
```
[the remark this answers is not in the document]
m5signorini: Previous to this, `FData` had no `cov` method. The signature of `FDataGrid` does not include `eval_points`, it is exactly `def cov(self: T) -> T:`.

m5signorini:
Latest commits were done in order to overload the `cov` method, adding the possibility to pass points of evaluation.
I believe that due to the nature of `@overload`, only the last method is used as implementation and taking into account the overloaded methods in the parent class is tricky, thus I needed to include the same signatures in both `FDataBasis` and `FDataGrid`.
I don't know if there is a better way to overload the method other than expecting optionally `None` values and checking at runtime.

skfda/inference/hotelling/_hotelling.py
```diff
@@ -85,7 +85,7 @@ def hotelling_t2(
     n = n1 + n2  # Size of full sample
     m = fd1.mean() - fd2.mean()  # Delta mean
 
-    if isinstance(fd1, FDataBasis):
+    if isinstance(fd1, FDataBasis) and isinstance(fd2, FDataBasis):
```
m5signorini, line 88: As of now, `cov()` returns and `FData` object and `cov(s,t)` returns the evaluation of the covariance in the grid formed by the cartesian product of s and t. Thus, there is no direct method for getting the data matrix or the coefficients matrix of the covariance function.

skfda/inference/anova/_anova_oneway.py
```diff
@@ -190,6 +190,7 @@ def _anova_bootstrap(
     random_state: RandomStateLike = None,
     p: int = 2,
     equal_var: bool = True,
+    default_n_features: int = 501,
```
m5signorini, line 193: Since currently, in case of dealing with a basis representation, the covariance is discretized, it is needed to specify the number of points of evaluation. Here I am using 501 points, same as in `to_grid`.

m5signorini:
Latest commits aim to eliminate the use of `cov()` and use instead `cov(s,t)` for evaluating the covariance function. This changes are applied only where the data matrix is needed. The only other instances of calls `cov()` appear in `CovarianceEstimator` and for now it is expected to use a `FData` object.

skfda/inference/anova/_anova_oneway.py
vnmabus: I think we should try to generate the coefficients directly. In any case, you should use a named constant instead of `501`.

skfda/tests/test_fdata_cov.py
```diff
@@ -0,0 +1,186 @@
...
+from skfda.datasets import make_gaussian_process
+from skfda.misc.covariances import CovarianceLike, Gaussian
+from skfda.representation import FDataBasis
+from skfda.representation.basis import Basis, BSplineBasis, FourierBasis
+
+############
+# FIXTURES
+############
+
+
```
vnmabus, line 16: Are all these fixtures really needed? Most of them are fixed to one value.

skfda/tests/test_fdata_cov.py
```diff
@@ -0,0 +1,186 @@
...
+    """Generate a mean."""
+    return request.param
+
+
+@pytest.fixture(
+    params=[
+        0.1,
+    ],
+)
+def noise(
```
vnmabus, line 101: Why do you test with noise? With noiseless data it is easier to use tests where you know the true answer.

vnmabus (changes requested):

skfda/ml/classification/_qda.py
```diff
@@ -207,7 +207,10 @@ def _fit_gaussian_process(
 
             cov_estimators.append(cov_estimator)
             means.append(cov_estimator.location_)
-            covariance.append(cov_estimator.covariance_.data_matrix[0, ..., 0])
+            # TODO: QDA should use the covariance estimators interface
```
m5signorini, line 210: Currently, QDA assumes that 'covariance_' is an 'FDataGrid'.
However, following this pull request, implementing RKHS and the Mahalanobis distance, will allow us to work only using the interface provided by the 'CovarianceEstimator' class.

This is simply a temporal fix until QDA is refactored, when adding the Mahalanobis distance.

skfda/inference/anova/_anova_oneway.py
```diff
@@ -203,26 +205,35 @@ def _anova_bootstrap(
             )
 
     # List with sizes of each group
-    sizes = [fd.n_samples for fd in fd_grouped]
+    sizes = [fdg.n_samples for fdg in fd_grouped]
 
     # Instance a random state object in case random_state is an int
     random_state = validate_random_state(random_state)
 
-    if equal_var:
```
m5signorini, line 211: Now, the `cov` method does not implicitly return a matrix, it needs to have the points of evaluation specified. Therefore, specifying a grid beforehand is required. However, in case of dealing with `FDataBasis`, there is no default grid. Thus, I added a default number of points of evaluation (features) in order to retain the same procedure to extract `grid_points`.

skfda/exploratory/stats/covariance/_parametric_gaussian.py
```diff
@@ -48,8 +53,14 @@ def fit(
         regressor.fit(grid_points, data_matrix.T)
 
         # TODO: Skip cov computation?
-        self.covariance_ = X.cov().copy(
-            data_matrix=regressor.kernel_(grid_points)[np.newaxis, ...],
+        # TODO: Use a user-public structure to represent the covariance,
+        #  instead of a Callable object
+        self.covariance_ = X.cov()
+        assert isinstance(self.covariance_, EmpiricalGrid)
+        self.covariance_.cov_fdata = self.covariance_.cov_fdata.copy(
```
m5signorini, line 60: Copying the `FData` allows to set the `data_matrix` correctly. Directly setting the attribute does not work, since the dimensions must be adjusted. Alternatively, one may do the same as implemented in the constructor of `FData`, which is calling `_int_to_real(np.atleast_2d(data_matrix))`.

skfda/representation/_functional_data.py
```diff
@@ -819,6 +820,51 @@ def sum(  # noqa: WPS125
 
         return self
 
+    @overload
+    def cov(
+        self: T,
+        s_points: NDArrayFloat,
```
vnmabus, line 826: As the minimum supported Python version is 3.8, I would set these parameters as positional-only.

vnmabus (changes requested):

skfda/representation/_functional_data.py
```diff
@@ -819,6 +820,52 @@ def sum(  # noqa: WPS125
...
+    @overload
+    def cov(
+        self: T,
+    ) -> Callable[[NDArrayFloat, NDArrayFloat], NDArrayFloat]:
+        pass
+
+    @abstractmethod
+    def cov(  # noqa: WPS320
+        self: T,
+        s_points: Optional[NDArrayFloat] = None,
```
vnmabus, line 841: The parameters should be positional only here too.

skfda/representation/basis/_fdatabasis.py
```diff
@@ -461,25 +463,57 @@ def var(self: T, eval_points: Optional[NDArrayFloat] = None) -> T:
...
+
+    @overload
+    def cov(
+        self: T,
+    ) -> Callable[[NDArrayFloat, NDArrayFloat], NDArrayFloat]:
+        pass
+
+    def cov(  # noqa: WPS320
+        self: T,
+        s_points: Optional[NDArrayFloat] = None,
```
vnmabus, line 483: And here also, positional only.

skfda/representation/grid.py
```diff
@@ -593,44 +595,51 @@ def var(self: T) -> T:
...
+
+    @overload
+    def cov(
+        self: T,
+    ) -> Callable[[NDArrayFloat, NDArrayFloat], NDArrayFloat]:
+        pass
+
+    def cov(  # noqa: WPS320
+        self: T,
+        s_points: Optional[NDArrayFloat] = None,
```
vnmabus, line 615: Here positional only too.

vnmabus (changes requested):

vnmabus (approved):
LGTM

PR #519, opened 2023-03-01
Speed up the calculation of the penalization matrix for FDataGrid

Ddelval:
Taking advantage of the sparsity of the penalization matrix, it is possible to improve the time complexity of its calculation significantly


skfda/misc/operators/_linear_differential_operator.py
```diff
@@ -650,4 +734,21 @@ def custombasis_penalty_matrix_optimized(
...
     """Optimized version for CustomBasis."""
+    if isinstance(basis.fdata, FDataGrid):
+        product_matrix = np.empty((basis.n_basis, basis.n_basis))
+        operator_evaluated = linear_operator(basis.fdata)(
+            basis.fdata.grid_points[0],
+        )
+
+        # Discard last dimension
+        operator_evaluated = operator_evaluated[..., 0]
+        for i in range(basis.n_basis):
```
vnmabus, line 745: I think that this can be easily vectorized. In fact, is it not possible to use the function `inner_product_matrix` for this purpose?

skfda/misc/operators/_linear_differential_operator.py
```diff
@@ -585,42 +635,76 @@ def fdatagrid_penalty_matrix_optimized(
...
+
+    # Formula derived from the definition of coefficients_not_uni
+    # in findiff.
+    spread = 2 * math.floor((max_deriv + 1) / 2) - 1
+    # The accuracy parameter in findiff increases the number of
+    # points considered to the left and the right
+    spread += findiff_accuracy + 1
+
+    product_matrix = np.zeros((basis.n_samples, basis.n_samples))
+    for i in range(n_points):
```
vnmabus, line 670: This logic is very complicated. I also feel that is repeating nontrivial work that was already done in `_optimized_operator_evaluation_in_grid`.

vnmabus (changes requested):

skfda/misc/operators/_linear_differential_operator.py
```diff
@@ -577,6 +579,54 @@ def bspline_penalty_matrix_optimized(
...
+    """
+    n_points = grid_points.shape[0]
+    diff_coefficients = np.zeros((n_points, n_points))
+
+    # The desired matrix can be obtained by appending the coefficients
+    # of the finite differences for each point column-wise.
+    for dif_order, w in enumerate(linear_operator.weights):
+        if w == 0:
+            continue
+        for point_index in range(n_points):
```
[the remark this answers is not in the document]
Ddelval: It is not slow. On the contrary, this whole function only takes around 3% of the runtime spent calculating the penalty matrix. I understand loops usually indicate slow code, but, in this case, having a linear algorithm outweighs this issue. 
Regarding the vectorization, I did look for it in the documentation but to no avail. Moreover, since there are no performance issues, I do not see a need for it. They are also unlikely to add it. It is already possible to convert a differential operator (as an object in findiff) to a matrix, so they would probably point to that. However, I tested it, and it was considerably slower.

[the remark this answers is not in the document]
Ddelval: This is the relevant portion of the call tree in the profiling output. The whole running time corresponds to executing the last test case of the FPCA regularization test but with a grid of 5000 points. As you can see, the running time of this function amounts to a negligible portion of the total running time.
image

skfda/misc/operators/_linear_differential_operator.py
```diff
@@ -585,42 +635,76 @@ def fdatagrid_penalty_matrix_optimized(
...
+    # maximum number of points to the left and right of the
+    # point where the linear operator is applied that can be
+    # different from zero.
+
+    # Formula derived from the definition of coefficients_not_uni
+    # in findiff.
+    spread = 2 * math.floor((max_deriv + 1) / 2) - 1
+    # The accuracy parameter in findiff increases the number of
+    # points considered to the left and the right
+    spread += findiff_accuracy + 1
```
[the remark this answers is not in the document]
Ddelval: I don't understand the question. Could you expand on it?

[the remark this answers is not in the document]
vnmabus: I think that the spread can be computed from the offsets already computed, without repeating the internal logic of `findiff` here.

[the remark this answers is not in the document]
Ddelval: Oh, I didn't think of that. I think it should be possible and it would make it safer in terms of making sure that this code is still applicable even if that internal logic changes. I'll change it and test it.

skfda/misc/operators/_linear_differential_operator.py
Ddelval: We could probably discuss this in more detail this week since I do not see exactly how to vectorize it. It is not equivalent to `inner_product_matrix` since when calling `inner_product_matrix`, the grid in which the function is evaluated is not specified. And, in the testing I have performed, that makes a huge difference. Maybe that is because I have tested with very fine grids. However, since the outer loops will have few iterations, especially compared to the integrations across all discretization points, I did not think that the loops were an issue.

skfda/misc/operators/_linear_differential_operator.py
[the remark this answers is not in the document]
Ddelval: Unfortunately, that is not possible. It took me a bit to remember why, but the issue is that the matrix is constructed point by point. That is to say, if you look at the final matrix returned by the function (it is transposed at the end), the offsets will tell you how many elements are not 0 column-wise, not row-wise.

skfda/misc/operators/_linear_differential_operator.py
vnmabus: Can't you construct a `FDataGrid` with that grid and call then `inner_product_matrix`?

Ddelval: Maybe that could work. I'll have a look when I have some more time (I have a couple of assignments due this week)

Ddelval: Actually yes. That works great. I have changed it in bb059682758b1562a298943ed775fcef551804d3. Thanks!

skfda/misc/operators/_linear_differential_operator.py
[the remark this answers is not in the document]
Ddelval: Nevermind. It is possible. It's not entirely straightforward, and I couldn't find a way to do it initially. I now have a solution for this and I have implemented it in 6d60b05417cd85aec81439a78711f555cc975272.

skfda/misc/operators/_linear_differential_operator.py
Ddelval: The new logic should be less complicated. It should be readable now. I think the complexity might have moved into `_optimized_operator_evaluation_in_grid`. I tried to mitigate it with some comments, as the meaning of the coefficients is not as straightforward as I would like. 
The issue is that the coefficients for the $i$-th point are intended to give you a way to calculate the value of the operator at that point. This is a similar problem to calculating the result of applying the operator to the delta basis. However, it is not entirely immediate (at least for me) that concatenating the coefficients column-wise and reading $j$-th row of the resulting matrix leads to the valuex of applying the operator to the $j$-th function of the delta basis.

skfda/representation/basis/_grid_basis.py
```diff
@@ -0,0 +1,68 @@
...
+    def __repr__(self) -> str:
+        return (
+            f"{type(self).__name__}("
+            f"grid_points={self.grid_points}, "
+        )
+
+    def __hash__(self) -> int:
+        return (
+            super().__hash__()
+            ^ hash(
```
vnmabus, line 65: XOR?

skfda/representation/basis/_grid_basis.py
```diff
@@ -0,0 +1,68 @@
...
+                self.grid_points,
```
vnmabus, line 66: I do not think this is hashable.

skfda/misc/operators/_linear_differential_operator.py
```diff
@@ -577,50 +579,188 @@ def bspline_penalty_matrix_optimized(
...
+    findiff_accuracy: int,
+) -> Tuple[NDArrayFloat, NDArrayInt]:
+    """
+    Compute the linear operator applied to the delta basis of the grid.
+
+    The delta basis is the basis where the i-th basis function is 1 in the
+    i-th point and 0 in the rest of points.
+
+    Returns:
+     - A matrix where the i-th row contians the values of the linear operator
```
vnmabus, line 594:
```suggestion
     - A matrix where the i-th row contains the values of the linear operator
```

skfda/misc/operators/_linear_differential_operator.py
```diff
@@ -577,50 +579,188 @@ def bspline_penalty_matrix_optimized(
...
+    domain_limits = np.hstack(
+        [
+            np.ones((n_points, 1)) * n_points,
+            np.zeros((n_points, 1)),
+        ],
+    ).astype(int)
+
+    # The desired matrix can be obtained by appending the coefficients
+    # of the finite differences for each point column-wise.
+    for dif_order, w in enumerate(linear_operator.weights):
```
vnmabus, line 647: You should check `linear_operator.constant_weights` as the weights could be callables, in principle.

vnmabus (changes requested):

skfda/representation/basis/_grid_basis.py
```diff
@@ -0,0 +1,60 @@
...
+        return (
+            super().__eq__(other)
+            and self.grid_points == other.grid_points
+        )
+
+    def __repr__(self) -> str:
+        return f"{type(self).__name__}(grid_points={self.grid_points}) "
+
+    def __hash__(self) -> int:
+        return hash((super(), f"{self.grid_points}"))
```
vnmabus, line 60: Is `grid_points` hashable?

vnmabus (changes requested):

skfda/representation/basis/_grid_basis.py
Ddelval: @vnmabus,
By itself no. But converted to a string it is no longer mutable so I would hope so. I might be missing something but that's why it's in an f-string.

vnmabus: Uh-uh, that is a bit fishy. I would convert it to tuples instead or something like that (see what `BSplineBasis` does with the `knots`).

In the long term, I think that the `hash` for the bases should not depend on the attributes, as the class is mutable.

Ddelval: I would consider that changing the `grid_points` results in a different grid basis. That is why I included it in the hash. I think it defines the identity of the basis.  
I have changed it to use a tuple instead of convertin to string

vnmabus: That is true. However, we only need that objects that compare equal have the same hash. The opposite, although desirable for performance reasons, is not required. Note that our only reason for having hashable bases in the first case is because they are an attribute for the `FDataBasis` ExtensionDType, and pandas requires that those are hashable.

I think that in the long term we should just hash the class type itself or something like that.

vnmabus: For dictionaries, only the keys need to be hashable. Normally, mutable objects such as lists or arrays are not made hashable on purpose.

Ddelval: Yeah, you are right. I was also thinking of sets, and I got them mixed up. 
To me, `GridBasis` does seem like it should be an immutable object. And it is fully characterized by the discretization points. With that said, since we do not have a need to implement effective hash methods, I think it is not unreasonable just to use the basis type as a hash. But then the method would be basically returning a constant. 

Personally, I'd rather change the `grid_points` to a property, making the object immutable, and keep the points in the hash. But I think the other option also makes sense. In any case, let me know what you think because I want to get to closing this PR rather sooner than later. The PLS functionality depends on this, and I'd like to have a clean PLS  PR at this point (with these commits integrated into develop).

vnmabus: I think I will accept this PR in its current state. If we change the hash behaviour for basis objects, that should be done in another PR.

For me, at the beginning I though of basis objects as immutable too. However, they were several disadvantages with that approach, and I am reconsidering it now. First, Python does not have first-class support for custom immutable objects, apart from frozen dataclasses. Second, and more important, the current approach makes difficult or even impossible the choice of basis parameters in a `BasisSmoother` transformer using hyperparameter selection tools such as `GridSearchCV`. For example, for choosing the number of elements in the basis, one has to create different `Basis` objects with the appropriate parameter set. Choosing several basis parameters at the same time is too complex for average users.

Forgetting the idea of immutable basis, we can make `Basis` inherit from `BaseEstimator`, thus allowing it to play well with the hyperparameter selection tools in scikit-learn. I think we should investigate this approach further.

vnmabus (approved):

PR #523, opened 2023-03-20
Update THANKS.txt

jltorrecilla:
JL update and minor mistakes


vnmabus (approved):
LGTM. In the future we should probably look at alternatives for keeping the list of contributors up-to-date, such as https://allcontributors.org/.

PR #531, opened 2023-03-31
Feature/concurrent regression

rafa9811:
Functional response with multivariate and concurrent covariates. Some fixes in linear regression with scalar response. Tests and examples.


vnmabus (approved):
I think the better approach with this PR is to merge it, as the functionality works, and then have a look to the code myself, in order to have a global vision and simplify things.

Issue #530, opened 2023-03-31
SRSF doc wrong formula


ego-thales:
Hello,

I just wanted to point out that the [SRSF doc](https://fda.readthedocs.io/en/latest/modules/misc/autosummary/skfda.misc.operators.SRSF.html) defines
$$SRSF(f) := \text{sign}(f)\sqrt{\vert f'\vert},$$
instead of (notice the derivative in the sign function)
$$SRSF(f) := \text{sign}(f')\sqrt{\vert f'\vert}.$$

Cheers

ego-thales:
_Edited the issue since the code isn't bugged as I thought it might be at first glance._

vnmabus:
Thank you for reporting it! Do you want to provide a fix?

ego-thales:
Ok I will, but probably later next week :+1:

vnmabus:
Solved in #542.

PR #542, opened 2023-06-16
SRSF formula doc fix

ego-thales:
See #530.

*Sorry for the delay*

### DIFF
--- a/skfda/misc/metrics/_fisher_rao.py
+++ b/skfda/misc/metrics/_fisher_rao.py
@@ -61,8 +61,8 @@ class FisherRaoDistance():
 
     .. math::
         d_{FR}(f_i, f_j) = \| q_i - q_j \|_2 =
-        \left ( \int_0^1 sgn(\dot{f_i}(t))\sqrt{|\dot{f_i}(t)|} -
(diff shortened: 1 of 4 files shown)


vnmabus:
> _Sorry for the delay_

You do not have to apologize. This is open source software, done in our free time, and nobody is entitled to our unpaid labor.

Thank you for the correction in the docs. However, I would say that the Newton notation for the derivative is not very widely used outside of the area of Differential Geometry. Thus, I would add a sentence such as "where $\dot{f}$ is the first derivative of $f$" or something like that. What do you think?

ego-thales:
> I would say that the Newton notation for the derivative is not very widely used outside of the area of Differential Geometry. Thus, I would add a sentence such as "where f˙ is the first derivative of f" or something like that. What do you think?

I agree completely. Personally, I would directly change all dots to apostrophes. I only tried to keep consistency. Let me know, I can update the PR.

vnmabus:
I think that would be great, thank you!

ego-thales:
> I think that would be great, thank you!

There you go :ok_hand:

vnmabus (approved):
LGTM

PR #546, opened 2023-06-22
Change type of axis in FData's mean method to accept zero.

Ddelval:
As the documention specifies, axis can be either None or 0.
Currently, if 0 is given as the axis, a mypy error is raised since the function signature only accepts a None axis.

### DIFF
--- a/skfda/representation/_functional_data.py
+++ b/skfda/representation/_functional_data.py
@@ -822,7 +822,7 @@ class FData(  # noqa: WPS214
     def mean(
         self: T,
         *,
-        axis: None = None,
+        axis: Optional[int] = None,
         dtype: None = None,
         out: None = None,
         keepdims: bool = False,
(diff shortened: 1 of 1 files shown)


skfda/representation/_functional_data.py
```diff
@@ -822,7 +822,7 @@ def sum(  # noqa: WPS125
     def mean(
         self: T,
         *,
-        axis: None = None,
+        axis: Optional[int] = None,
```
vnmabus, line 825: Use `int | None` instead.

vnmabus (changes requested):

skfda/representation/_functional_data.py
Ddelval: I used the other notation to be coherent with the rest of the file. 
Changed now.

vnmabus (approved):
LGTM

PR #550, opened 2023-06-25
RKHS inner product

m5signorini:
This PR adds the RKHS inner product.

Cases regarding the product between `FDataBasis` objects with different basis or with the covariance function not already expressed in the tensor basis, might be better to not consider them.

### DIFF
--- a/skfda/misc/covariances.py
+++ b/skfda/misc/covariances.py
@@ -804,6 +804,8 @@ class Empirical(Covariance):
         Returns:
             Covariance function evaluated at the grid formed by x and y.
         """
+        x = _transform_to_2d(x)
+        y = _transform_to_2d(y)
         return self.cov_fdata([x, y], grid=True)[0, ..., 0]
 
 
--- a/skfda/misc/rkhs_product.py
+++ b/skfda/misc/rkhs_product.py
@@ -139,13 +139,6 @@ def rkhs_inner_product(
             )[0],
         ).T[..., np.newaxis]
 
-        # NOTE: Numpy alternative.
-        # inv_fd2_matrix = np.linalg.lstsq(
-        #     cov_matrix_1_2.T,
-        #     data_matrix_2[..., 0].T,
-        #     rcond=None,
-        # )[0].T[..., np.newaxis]
-
     products = (
         np.transpose(data_matrix_1, axes=(0, 2, 1))
         @ inv_fd2_matrix
--- a/skfda/misc/rkhs_product.py
+++ b/skfda/misc/rkhs_product.py
@@ -1,0 +1,223 @@
+"""RKHS inner product of two FData objects."""
+from __future__ import annotations
+
+from typing import Callable, Optional
+
+import multimethod
+import numpy as np
+import scipy.linalg
+
+from ..misc.covariances import EmpiricalBasis
+from ..representation import FData, FDataBasis, FDataGrid
+from ..representation.basis import Basis, TensorBasis
+from ..typing._numpy import NDArrayFloat
+from ._math import inner_product
+from .validation import check_fdata_dimensions
+
+
+def _check_compatible_fdata(
+    fdata1: FData,
+    fdata2: FData,
+) -> None:
+    """Check terms of the inner product are compatible.
+
+    Args:
+        fdata1: First functional data object.
+        fdata2: Second functional data object.
+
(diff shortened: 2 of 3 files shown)


skfda/misc/covariances.py
```diff
@@ -804,6 +804,8 @@ def __call__(self, x: ArrayLike, y: ArrayLike) -> NDArrayFloat:
         Returns:
             Covariance function evaluated at the grid formed by x and y.
         """
+        x = _transform_to_2d(x)
+        y = _transform_to_2d(y)
```
m5signorini, line 808: Fixed for consistency, it was missing from #505 .

skfda/misc/rkhs_product.py
```diff
@@ -0,0 +1,216 @@
...
+    else:
+        inv_fd2_coefs = np.linalg.lstsq(
+            cov_coeff_matrix.T,
+            fdata2.coefficients.T,
+            rcond=None,
+        )[0]
+
+    # Following einsum is equivalent to doing the matrix multiplication
+    # and then taking the diagonal of the result
+    return np.einsum(  # type: ignore[no-any-return]
```
vnmabus, line 212: Is this faster/more memory efficient than the solution in https://stackoverflow.com/a/14759273/2455333 ? Just out of curiosity.

vnmabus (changes requested):

skfda/misc/rkhs_product.py
m5signorini: I used `cProfile` to test both methods with 10000 x 10000 matrices and `einsum` is significantly (~14 times) faster. This is mainly because of the `dot` method, since it calculates all products instead of only the diagonal ones.

m5signorini:
The test errors occur in `test_pandas_fdatabasis` and `test_pandas_fdatagrid` .
They say that the fixture `all_boolean_reductions` can not found.

skfda/misc/rkhs_product.py
```diff
@@ -0,0 +1,256 @@
...
+            cov_matrix_1_2,
+            data_matrix_2,
+        )
+    else:
+        inv_fd2_matrix = np.asarray(
+            scipy.linalg.lstsq(
+                cov_matrix_1_2.T,
+                data_matrix_2[..., 0].T,
+                cond=cond,
+            )[0],
```
m5signorini, line 173: I find it hard to replace this transpositions because the parameters must be of dimensions (M,N) and (M,K) respectively. Since `cov_matrix` has dimensions (grid_points_of_fdata1, grid_points_of_fdata2) and `data_matrix` has dimensions (n_samples, grid_points, 1), in order to match dimensions both transpositions are required.

skfda/misc/rkhs_product.py
```diff
@@ -0,0 +1,256 @@
...
+        ).T[..., np.newaxis]
```
m5signorini, line 174: The returned value will be a matrix of dimensions **(N,K)** which corresponds to **(n_grid_points_fdata1, n_samples_fdata2)**.

Meanwhile, `np.linalg.solve`  expects dimensions **(..., M, M)** and **(..., M, K)**, which are **(n_grid_points_fdata1, n_grid_points_fdata2)** and **(n_samples_fdata2, n_grid_points_fdata2, 1)**.
The output will then be of dimensions **(..., M, K)**, which are  **(n_samples_fdata2, n_grid_points_fdata1, 1)**.

This means that to be coherent with `np.linalg.solve` the resulting array from `lstsq` of dimensions **(n_grid_points_fdata1, n_samples_fdata2)** has to be rearranged as an array of dimensions **(n_samples_fdata2, n_grid_points_fdata1, 1)**, thus the transposition and new axis.

I fail to see a simpler method at the moment that doesn't involve changing the rest, I'll keep thinking in any case.

vnmabus (dismissed):
LGTM. Should it be merged?

vnmabus:
@m5signorini Maybe I wasn't clear. I was asking if there is anything else that you wanted to include, as the PR is marked as a draft, or if I have your permission to merge it.

m5signorini:
Yes sorry, forgot to change it from draft. I think it can be merged.

vnmabus:
Thank you for your answer! However, a test is failing (and the difference between the expected and actual result is too high). Can you take a look at it?

PR #552, opened 2023-06-30
Polish fpca regression example

Ddelval:
Several issues have been addressed:
- Fix the labels of the cross validation plot
- Change cross validation plot style
- Round scores to four decimal places


Ddelval:
Blocked until tests are fixed ( #553 )

vnmabus (changes requested):
Please store the results in a variable before printing (do not make f-strings that complicated).

vnmabus (approved):

PR #556, opened 2023-07-12
Feature/555 ddof parameter in fdatacov

### LINKED ISSUE #555: ddof parameter in FData.cov
The current behavior of the FData.cov method is to normalize the covariance by $N-1$ (default normalization applied by [`np.cov`](https://numpy.org/doc/stable/reference/generated/numpy.cov.html)). This limits the functionality of the covariance function. In order to provide a more flexible covariance method, a parameter should be added so that this normalization coefficient can be specified by the user.

**Describe the solution you'd like**
Add a parameter `ddof` to FData.cov (by default `ddof=1`) such that the normalization function applied by FData.cov is `N-ddof`.

> ddof: int, optional
>    “Delta Degrees of Freedom”: the divisor used in the calculation is N - ddof, where N represents the number of elements. By default ddof is 1.

pcuestas:
Add parameter `ddof` to FData.cov, FData.var, stats.cov and stats.var.

As explained in #555, this `ddof` parameter has value 1 by default and means "Delta Degrees of Freedom". The divisor used in the calculation of the covariance is `N-ddof`.

This changes the previous default behavior of `var`, which used `ddof=0` by default. Therefore, changes have been applied to scikit-fda/skfda/misc/scoring.py (Line 77) and scikit-fda/skfda/ml/clustering/_kmeans.py (Line 150), in order to preserve the previous behaviour of those modules.

This pull request closes #555 .

### DIFF
--- a/skfda/exploratory/stats/_stats.py
+++ b/skfda/exploratory/stats/_stats.py
@@ -41,19 +41,22 @@ def mean(
     return (X * weight).sum()
 
 
-def var(X: FData) -> FDataGrid:
+def var(X: FData, ddof: int = 1) -> FDataGrid:
     """
     Compute the variance of a set of samples in a FData object.
 
     Args:
         X: Object containing all the set of samples whose variance is desired.
+        ddof: "Delta Degrees of Freedom": the divisor used in the calculation
+            is `N - ddof`, where `N` represents the number of elements. By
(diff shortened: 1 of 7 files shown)


vnmabus (approved):
LGTM.

Issue #560, opened 2023-08-10
New `FDataGrid.restrict` option `include_bounds`?


ego-thales:
Hi,

Currently, `FDataGrid.restrict(domain)` eliminates points outside of `domain`. As a consequence, the first (*resp.* last) grid point (in the 1D case) might be relatively far from the lower (*resp.* upper) bound of `domain`. In this case, **after** restriction, between the bound and the grid point, `skfda` will essentially use extrapolation, when only interpolation should occur, which might lower the accuracy of the "guess".

For example:
![tmp](https://github.com/GAA-UAM/scikit-fda/assets/121242234/62e4fb42-50b0-4be1-866d-bbf2c27e3b66)

This is why I propose to add an `include_bounds` option, that would interpolate at boundaries before restricting if the grid points don't include the `domain` values. The option default value should be discussed. If `True` might seem better, it will also silently add new grid_points, which might not be wanted.

What do you think about it?
Élie

vnmabus:
I thought a little bit about it when implementing the restrict method. I agree that this problem exists, specially for measurements which are not very dense. At that moment I did not thought about interpolating at the boundaries, which is probably a reasonable approach. I thus would love to add such an option.

I think that this option as proposed would also be applicable to high dimensional functions (surfaces and so on). If that is true I have nothing against it.

vnmabus:
Solved in #561.

Issue #498, opened 2022-11-22
Data structure for discretized sparse data


vnmabus:
**Is your feature request related to a problem? Please describe.**
We need a `FData` subclass capable of storing discretized functions where:
- The points in which the functions are measured are NOT in a grid (specially relevant for high dimensional functions such as surfaces).
- The points in which each function is measured are different, and probably a different number of points per observation.
- There are typically few points measured per observation (sparse instead of dense).

This structure is necessary for efficient storage and computation with this kind of data.

**Describe the solution you'd like**
The proposal for the implementation is to have three arrays:
- A first array containing the (concatenated) input points for the observations. For example, for three functional observations $f_1$, $f_2$, $f_3$, evaluated at $f_1(1, 2) = (3, 4, 5), f_1(2, 4) = (1, 7, 3), f_2(0, 0) = (1, 0, 1), f_3(0, 1) = (2, 2, 2), f_3(3, 5) = (1, 0, 0)$, this array would contain `[(1, 2), (2, 4), (0, 0), (0, 1), (3, 5)]`.
- A second array with the corresponding values. In the previous example it would be `[(3, 4, 5), (1, 7, 3), (1, 0, 1), (2, 2, 2), (1, 0, 0)]`.
- A third array with the indexes where the points of each observation start. In this case `[0, 2, 3]`.

This approach has a compact representation and also allows for vectorization to be applied. The indexes can be used to apply the `reduceat` method of NumPy ufuncs.

**Describe alternatives you've considered**
A more high-level API should be also exposed and used when possible.

vnmabus:
After an internal discussion, and noticing that classes for representing irregular data are in the packages Tidyfun, fundata and fdaPy, we think that this class should definitely be implemented, but the name should reflect that is a class for representing irregular data. Sparsity is more a property of the actual dataset rather than of the structure.

Thus, @opintosant, you can start implementing this class, focusing on the evaluation and plot methods first.

vnmabus:
This was implemented in #536.

PR #561, opened 2023-08-11
`FDataGrid.restrict` option `with_bounds`

ego-thales:
Implemented proposed enhancement from #560.

A few notes to keep in mind when (or before) merging:
- `with_bounds` defaults to `True`,
- With high `dim_domain`, `domain`'s boundary potentially contains a lot of points, which might end up pretty slow. That said, I guess it has to happen when manipulating a functional object on a huge domain size,
- No 1D optimization was made. I don't know the general spirit of your implementation, but if you generally treat 1D differently with lighter/faster code, then this should also be done here.

Waiting for your feedback!

Cheers :)
Élie

### DIFF
--- a/skfda/representation/grid.py
+++ b/skfda/representation/grid.py
@@ -1113,7 +1113,7 @@ class FDataGrid(FData):  # noqa: WPS214
             ib = np.searchsorted(grid_points, b, 'right')
             slice_list.append(slice(ia, ib))
         grid_points = [g[s] for g, s in zip(self.grid_points, slice_list)]
-        data_matrix = self.data_matrix[slice(None), *slice_list]
+        data_matrix = self.data_matrix[(slice(None),) + tuple(slice_list)]
 
         # Ensure that boundaries are in grid_points.
         if with_bounds:
@@ -1122,10 +1122,10 @@ class FDataGrid(FData):  # noqa: WPS214
                 dim_points = grid_points[dim]
                 grid_points[dim] = []
                 add_low, add_high = False, False
-                if a < g:
+                if a < dim_points[0]:
                     add_low = True
                     grid_points[dim].append(a)
-                if b > gb:
+                if b > dim_points[-1]:
                     add_high = True
                     grid_points[dim].append(b)
                 if not (add_low or add_high):
@@ -1142,19 +1142,19 @@ class FDataGrid(FData):  # noqa: WPS214
                 if add_low:
                     if add_high:
                         data_matrix = np.concatenate(
-                            (edge_values.take(0, dim),
+                            (edge_values.take([0], axis=dim+1),
                              data_matrix,
-                             edge_values.take(1, dim)),
-                            axis=dim)
+                             edge_values.take([1], axis=dim+1)),
+                            axis=dim+1)
                         dim_points = np.concatenate(([a], dim_points, [b]))
                     else:
                         data_matrix = np.concatenate(
-                            (edge_values, data_matrix), axis=dim)
+                            (edge_values, data_matrix), axis=dim+1)
                         dim_points = np.concatenate(([a], dim_points))
                 else:
                     if add_high:
                         data_matrix = np.concatenate(
-                            (data_matrix, edge_values), axis=dim)
+                            (data_matrix, edge_values), axis=dim+1)
                         dim_points = np.concatenate((dim_points, [b]))
                 grid_points[dim] = dim_points
 
--- a/skfda/representation/grid.py
+++ b/skfda/representation/grid.py
@@ -1108,7 +1108,7 @@ class FDataGrid(FData):  # noqa: WPS214
         # Eliminate points outside the new range.
         index_list = []
         slice_list = []
-        for (a, b), grid_points in zip(domain_range,self.grid_points):
+        for (a, b), grid_points in zip(domain_range, self.grid_points):
             ia = np.searchsorted(grid_points, a)
             ib = np.searchsorted(grid_points, b, 'right')
             slice_list.append(slice(ia, ib))
--- a/skfda/representation/grid.py
+++ b/skfda/representation/grid.py
@@ -1082,12 +1082,16 @@ class FDataGrid(FData):  # noqa: WPS214
     def restrict(
         self: T,
         domain_range: DomainRangeLike,
+        *,
+        with_bounds: bool = True,
     ) -> T:
         """
         Restrict the functions to a new domain range.
 
         Args:
             domain_range: New domain range.
+            with_bounds: Whether or not to ensure domain boundaries
+                appear in `grid_points`.
 
         Returns:
             Restricted function.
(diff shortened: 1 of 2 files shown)


ego-thales:
I now notice that this option potentially modifies the result of every `FDataGrid.integrate` call with a `domain` argument. This probably explains the failed tests.

One easy solution would be to set default value to `False`.
The other option is to go through every failed test and check whether or not the difference can indeed be due to the more precise calculus.

In turn, it might be necessary to change some doc example, I'm not sure about it.

In any case, I will wait for your feedback and insights regarding this, bcause you have a much deeper view of the whole package.
:)

vnmabus:
The "proper" way to do it would be:
- Decide if we want that the user can include the bounds as new points. As far as I see it, the main advantage is increased accuracy in the evaluation, while the main issue is that we add points that were not directly measured. Both behaviors may be surprising.
- Note that there is also a third option (keep ALL points, just change the `domain_range` attribute) which could in principle be implemented. I think that currently the `FDataGrid` constructor disallows having grid points outside the domain range, but this invariant is never used AFAIK, so it could be removed. This approach keeps the original accuracy and does not add "synthetic" points, but it is probably wasteful, as most points are no longer needed.
- Thus, we need to discuss which options could be selected by the user, and if a `bool` is enough for that choice or we need an enum or a `Literal` parameter.
- We would probably want to define this method in a way that can be extended in the future to other `FData` subclasses such as `FDataBasis` and the future `FDataIrregular` class. For the former these options make no sense, but it make sense for the latter, so it would be better to take that into account from the beginning.
- Once we have decided the possible choices, we have to pick a default one, looking at the pros and cons of each.
- If the default option is different from the current one, we ideally should keep the old default with a deprecation warning for a while (requesting the user to explicitly choose an option to remove the warning). If we see that this method has no users outside the package, we can skip this. We could in principle do breaking changes as we are still in 0.x version, but it is not very polite.
- We have to check the internal uses of this method to choose the more appropriate option for each.

eliegoudout:
Hi,
Just a quick answer from my personal phone regarding other possible ways to do this.

I will argue **against** keeping outside points for 2 reasons:
- It will make interpolation vs. extrapolation blurry. For example, how to you extrapolate at `x = -.09`, with domain `[0, 1]` and grid points `-.1, .1, .3`? If you use `-.1` value, then you're much closer to interpolation than extrapolation, and if you don't, you lose precision.
- The general idea of "restricting" **IS**, imo, forgetting about outside information. Imagine wanting to test out an extrapolation method, then it is important that restriction doesn't know about outside values.

I agree that using `Tru` as default value might not be polite. I thus think that if we end up deciding that it would be the best option, then the way to go surely is warning users during the next release (or more), before applying the change.

Is there a way to know more about `FDataIrregular`?


I think a good way to choose the defaul value will be to test the accuracy gains on plausible cases, as well as execution time lost.

vnmabus:
>     * It will make interpolation vs. extrapolation blurry. For example, how to you extrapolate at `x = -.09`, with domain `[0, 1]` and grid points `-.1, .1, .3`? If you use `-.1` value, then you're much closer to interpolation than extrapolation, and if you don't, you lose precision.

That is true, although it is more a semantic issue. By default, the extrapolation strategy of `FDataGrid` delegates to interpolation. If you change it to something else, it would be applied to the points outside the domain, as usual.

>     * The general idea of "restricting" **IS**, imo, forgetting about outside information. Imagine wanting to test out an extrapolation method, then it is important that restriction doesn't know about outside values.

Extrapolation would be applied for the points outside the domain, in spite of having values measured for them. The default extrapolation, of course, uses these values.

> Is there a way to know more about `FDataIrregular`?

Yes, it is a class to represent functions measured at different points (not a fixed grid). It was first proposed in #498 and a tentative implementation is in #536.

> I think a good way to choose the defaul value will be to test the accuracy gains on plausible cases, as well as execution time lost.

I want to discuss also the issue with the other people involved in the project, as they may provide more insights. They are currently on vacation but we will meet again in September.

skfda/representation/grid.py
```diff
@@ -1101,30 +1105,61 @@ def restrict(
...
+                    else:
+                        data_matrix = np.concatenate(
+                            (edge_values, data_matrix), axis=dim+1)
+                        dim_points = np.concatenate(([a], dim_points))
+                else:
+                    if add_high:
+                        data_matrix = np.concatenate(
+                            (data_matrix, edge_values), axis=dim+1)
+                        dim_points = np.concatenate((dim_points, [b]))
+                grid_points[dim] = dim_points
```
vnmabus, line 1158: This code is a bit difficult to maintain. Please check if this suggestion does the same:
```suggestion
            # Proceed dim by dim
            for dim, (a, b) in enumerate(domain_range):
                left = []
                right = []

                if a < grid_points[dim][0]:
                    left.append([a])

                if b > grid_points[dim][-1]:
                    right.append([b])

                grid_points[dim] = np.concatenate(left + [grid_points[dim]] + right)
            data_matrix = self(grid_points, grid=True)
```

skfda/tests/test_grid.py
```diff
@@ -337,6 +337,18 @@ def test_evaluate_grid_unaligned(self) -> None:
 
         np.testing.assert_allclose(res, expected)
 
+    def test_restrict(self) -> None:
+        """ Test FDataGrid.restrict with bounds. """
+        # Test 1 sample function R^3 -> R^5.
+        grid_points = ([0, 1], [0, 1, 2], [0, 1, 2, 3])
+        data_matrix = np.ones((1, 2, 3, 4, 5))
+        fd = FDataGrid(data_matrix, grid_points)
+        fd_restricted = fd.restrict(((0, 1), (.5, 1.5), (.5, 2)))
```
vnmabus, line 346: Pass the `with_bounds` argument explicitly.

vnmabus (changes requested):

skfda/representation/grid.py
ego-thales: Oooh nice, I didn't know about the `grid` option of `FDataGrid.__call__`. This will certainly simplify things (as generating the evaluation mesh is no longer required).

On the other hand, I think your suggestion will reevaluate the function on *the whole domain* instead of on *the boundary*. If true, this is too costly to be considered and addition dimension by dimension should be preferred, imo. For example, with `dim_domain =3` and approximately `10` grid points per axis, your suggestion will evaluate `self` at `1000` points instead of only `60` being necessary.

I will incorporate the simplified syntax (with `grid` option) in the PR when I have the chance! I will also check how the `grid` option was implemented to ensure it is not slower (or not too much if at all) than the mesh approach.

vnmabus: It is true that it evaluates all the points. However, I am not sure if it will be noticeably slower, as most part of the computation would be vectorized, if I am not mistaken. Unless it is noticeably slower I think that I prefer this version as it seems more readable to me (and this function is unlikely to be in a bottleneck anyway).

skfda/tests/test_grid.py
```diff
@@ -337,6 +337,19 @@ def test_evaluate_grid_unaligned(self) -> None:
 
         np.testing.assert_allclose(res, expected)
 
+    def test_restrict(self) -> None:
+        """ Test FDataGrid.restrict with bounds. """
```
vnmabus, line 341:
```suggestion
        """Test FDataGrid.restrict with bounds."""
```

skfda/tests/test_grid.py
```diff
@@ -337,6 +337,19 @@ def test_evaluate_grid_unaligned(self) -> None:
...
+        # Test 1 sample function R^3 -> R^5.
+        grid_points = ([0, 1], [0, 1, 2], [0, 1, 2, 3])
+        data_matrix = np.ones((1, 2, 3, 4, 5))
+        fd = FDataGrid(data_matrix, grid_points)
+        restricted_domain = ((0, 1), (.5, 1.5), (.5, 2))
```
vnmabus, line 346:
```suggestion
        restricted_domain = ((0, 1), (0.5, 1.5), (0.5, 2))
```

skfda/tests/test_grid.py
```diff
@@ -337,6 +337,19 @@ def test_evaluate_grid_unaligned(self) -> None:
...
+        fd_restricted = fd.restrict(restricted_domain, with_bounds=True)
+        res = fd_restricted.grid_points
+        expected = ([0, 1], [.5, 1, 1.5], [.5, 1, 2])
```
vnmabus, line 349:
```suggestion
        expected = ([0, 1], [0.5, 1, 1.5], [0.5, 1, 2])
```

vnmabus, line 349: We should probably aid Mypy by explicitly typing this variable or using arrays instead of lists, as the type inferred for the tuple elements is `object`.

vnmabus (changes requested):

skfda/tests/test_grid.py
ego-thales: Which of the following two (if any) would you prefer?
```python
expected = map(np.array, ([0, 1], [0.5, 1, 1.5], [0.5, 1, 2]))
expected: tuple[list] = ([0, 1], [0.5, 1, 1.5], [0.5, 1, 2])
```

ego-thales: Second question then since it's too long:
```python
expected: Tuple[Sequence[float], ...] = ([0, 1],
                                         [0.5, 1, 1.5],
                                         [0.5, 1, 2])
# or
expected: Tuple[Sequence[float], ...] = (
    [0, 1],
    [0.5, 1, 1.5],
    [0.5, 1, 2],
)

```

vnmabus: The second one is the one accepted by [our style checker of choice](https://github.com/wemake-services/wemake-python-styleguide).

vnmabus (dismissed):
Just one silly suggestion by the style checker. The rest LGTM.

vnmabus (approved):
LGTM now.

Issue #541, opened 2023-06-13
Add standard deviation


vnmabus:
Add a method for computing standard deviation of functional data both in discretized and basis expansions.

pcuestas:
There is an issue regarding the design of the std function that should be specified, which is the normalization coefficient to be applied and whether it should be up to the user.

The definition of `std` provided in Kokoszka and Reimherr (2017) is:

$$(std_X(t) ) ^2=\frac{1}{N} \sum_{n=1}^{N} (X_n(t) - \bar{X}(t))^2.$$

This normalization by $N$ is the default used in [`numpy.var`](https://numpy.org/doc/stable/reference/generated/numpy.var.html) and [`numpy.std`](https://numpy.org/doc/stable/reference/generated/numpy.std.html), the latter being the most natural function to use when implementing `FDataGrid.std`:

```python
def std(X: FDataGrid) -> FDataGrid:
    return X.copy(
        data_matrix=np.array([np.std(X.data_matrix, axis=0)]),
        sample_names=("standard deviation",),
    )
```

However, the most natural way of defining the std of a FDataBasis object is to use the `FDataBasis.cov` method. `FDataBasis.cov` calculates the covariance using the formula:

$$(K_X(t, s) ) ^2=\frac{1}{N-1} \sum_{n=1}^{N} (X_n(t) - \bar{X}(t))(X_n(s) - \bar{X}(s)),$$

because $(N-1)$ is the default normalization used by [`numpy.cov`](https://numpy.org/doc/stable/reference/generated/numpy.cov.html).

A natural solution to this issue would be to make the signature of `std` similar to that of [`numpy.std`](https://numpy.org/doc/stable/reference/generated/numpy.std.html), where there is a parameter:
```
ddof: int, optional
    Means Delta Degrees of Freedom. The divisor used in calculations is N - ddof, where N represents the number of elements. By default ddof is zero.
```

pcuestas:
I closed this issue by accident. Sorry.

PR #566, opened 2023-08-17
Add standard deviation as a function

pcuestas:
Added `std` as a function using single dispatch in the `skfda.exploratory.stats` module.

The function takes two arguments: a `FData` object and a `ddof` (Delta Degrees Of Freedom) integer. By default, `ddof=1`. Rationale behind this decision has been that these functions (stats.cov, stats.var, stats.std) are meant to be used for statistical (exploratory) purposes, and not for Maximum Likelihood algorithms. The default is different from [`numpy.std`](https://numpy.org/doc/stable/reference/generated/numpy.std.html) (which uses `ddof=1`). [`pandas.DataFrame.std`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.std.html) and [`scipy.stats.variation`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.variation.html) use `ddof=1`.

The approach taken to calculate the std of a `FDataBasis` object is the more direct one that we discussed which does not use the covariance object. Instead, a matrix multiplication is performed to calculate the pointwise variance on an arbitrary point of the domain: 

https://github.com/GAA-UAM/scikit-fda/blob/6eada7287e7e1a6098553714cc9a6a06d7b34eef/skfda/exploratory/stats/_stats.py#L153-L156

Tests included to cover the general $\mathbb{R}^p \to \mathbb{R}^q$ case for both `FDataGrid` and `FDataBasis` objects.

Closes https://github.com/GAA-UAM/scikit-fda/issues/541.


skfda/exploratory/stats/_stats.py
```diff
@@ -99,6 +100,64 @@ def cov(
...
+        raise NotImplementedError(
+            "Standard deviation only implemented "
+            "for univariate functions.",
+        )
+
+    basis = X.basis
+    coeff_matrix = np.cov(X.coefficients, rowvar=False, ddof=ddof)
+
+    def std_function(t_points: NDArrayFloat) -> NDArrayFloat:
+        assert len(t_points) == 1, (
```
vnmabus, line 149: Would if be possible to lift this restriction? I want to check for more efficient quadrature routines in the future, that may allow evaluation at multiple points.

skfda/misc/_math.py
```diff
@@ -677,3 +677,33 @@ def cosine_similarity_matrix(
     return _clip_cosine(
         inner_matrix / norm1[:, np.newaxis] / norm2[np.newaxis, :],
     )
+
+
+def function_to_fdatabasis(
```
vnmabus, line 682: I would put it in `_utils` for now, as I am not sure that we want to expose it for final users at this point.

skfda/tests/test_stats.py
```diff
@@ -144,3 +155,48 @@ def test_mei(self) -> None:
             ]),
             rtol=1e-5,
         )
+
+
+class TestStd(unittest.TestCase):
+    """Test the standard deviation of fuctional data objects."""
+
+    def _test_std_gaussian_fourier(self, n_basis: int) -> None:
+        """Test standard deviation: gaussian processes and a Fourier basis."""
```
vnmabus, line 164: Adjectives that come from proper names are in uppercase in English.

```suggestion
        """Test standard deviation: Gaussian processes and a Fourier basis."""
```

skfda/tests/test_stats.py
```diff
@@ -144,3 +155,48 @@ def test_mei(self) -> None:
             ]),
             rtol=1e-5,
         )
+
+
+class TestStd(unittest.TestCase):
```
vnmabus, line 160: We are trying to go with Pytest-style tests for new tests, instead of unittest-style ones. Please take a look at them.

vnmabus (changes requested):

skfda/exploratory/stats/_stats.py
```diff
@@ -99,6 +100,64 @@ def cov(
...
+        data_matrix=np.std(X.data_matrix, axis=0, ddof=ddof)[np.newaxis, ...],
+        sample_names=("standard deviation",),
+    )
+
+
+@std.register(FDataBasis)
+def std_fdatabasis(X: FDataBasis, ddof: int = 1) -> FDataBasis:
+    from ...misc._math import function_to_fdatabasis
+
+    if X.dim_domain != 1 or X.dim_codomain != 1:
```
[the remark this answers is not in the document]
pcuestas: I have not seen a definition of std for other types of functional data objects. I guess a natural definition would be to define it as: 

For each $\textbf{t}$ in the domain, the $k$-th coordinate of $\text{std}(\textbf{t})$ is $std \left(\pi_k \left( x_i(\textbf{t})\right) \right) _{i=1}^N$. Where $\pi_k$ is the projection over the $k$-th coordinate of the codomain.

skfda/exploratory/stats/_stats.py
```diff
@@ -99,6 +100,64 @@ def cov(
...
+
+    """
+    raise NotImplementedError("Not implemented for this type")
+
+
+@std.register(FDataGrid)
+def std_fdatagrid(X: FDataGrid, ddof: int = 1) -> FDataGrid:
+    return X.copy(
+        data_matrix=np.std(X.data_matrix, axis=0, ddof=ddof)[np.newaxis, ...],
+        sample_names=("standard deviation",),
```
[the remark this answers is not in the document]
pcuestas: `sample_names` is required if nothing is changed.

By default, `X.copy()` sets the sample names of its output to `X.sample_names`, which has more length in general than 1 (number of samples of the std).

skfda/exploratory/stats/_stats.py
pcuestas: I think this is solved now. `dim_codomain` and number of points in `t_points` may be greater than 1.
https://github.com/GAA-UAM/scikit-fda/blob/7a8ece04f7bcc929eba3445d129abe4d7a54daf4/skfda/exploratory/stats/_stats.py#L146-L158

The computation of variance (line 154) can also be done with one numpy function call to `einsum`, but this is more explicit.
Alternative:
```python
np.einsum("bi,bd,di->i", basis_evaluation, coeff_cov_matrix, basis_evaluation)
```

skfda/exploratory/stats/_stats.py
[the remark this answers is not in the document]
pcuestas: After [adapting `std_function`](https://github.com/GAA-UAM/scikit-fda/pull/566#discussion_r1307642793) to FDataBasis of $\mathbb{R}^p \rightarrow \mathbb{R}^q$ functions, there are still issues when $p$ or $q$ is greater than 1.
* If $p>1$: the integration algorithm seems to take too long (more than 15 minutes).
* If $q>1$: the integration algorithm returns `nan` values in the coordinates different from the first. Besides, I think `function_to_fdatabasis` has to be modified.

skfda/exploratory/stats/_stats.py
```diff
@@ -99,6 +100,64 @@
...
+    if X.dim_domain != 1 or X.dim_codomain != 1:
+        raise NotImplementedError(
+            "Standard deviation only implemented "
+            "for univariate functions."
+        )
+
+    basis = X.basis
+    coeff_matrix = np.cov(X.coefficients, rowvar=False, ddof=ddof)
+
+    def std_function(t_points: NDArrayFloat) -> NDArrayFloat:
```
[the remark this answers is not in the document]
pcuestas: I think it is necessary

skfda/tests/test_stats_std.py
```diff
@@ -0,0 +1,207 @@
...
+) -> None:
+    """Test standard deviation: Gaussian processes and a Fourier basis."""
+    fourier_basis = FourierBasis(n_basis=n_basis, domain_range=(0, 1))
+    fd = gaussian_process.to_basis(fourier_basis)
+
+    std_fd = std(fd)
+    grid = np.linspace(start, stop, n_features)
+    almost_std_fd = std(fd.to_grid(grid)).to_basis(fourier_basis)
+
+    inner_grid_limit = n_features // 10
```
vnmabus, line 76: Honestly, I do not get what you are trying to do in this test.

skfda/tests/test_stats_std.py
```diff
@@ -0,0 +1,207 @@
...
+
+    outer_grid = grid[:inner_grid_limit] + grid[-inner_grid_limit:]
+    np.testing.assert_allclose(
+        std_fd(outer_grid),
+        almost_std_fd(outer_grid),
+        rtol=1e-2,
+    )
+
+
+@pytest.mark.parametrize("fdatagrid, expected_std_data_matrix", [
```
vnmabus, line 92: You are sacrificing readability in order to prevent the replication of an assert. I would split it in different tests, with proper comments indicating the particularities of each (e.g. dimension of the domain/codomain).

skfda/tests/test_stats_std.py
```diff
@@ -0,0 +1,207 @@
...
+    expected_std_data_matrix: NDArrayFloat,
+) -> None:
+    """Test some std_fdatagrid cases."""
+    np.testing.assert_allclose(
+        std(fdatagrid).data_matrix,
+        expected_std_data_matrix,
+    )
+
+
+@pytest.mark.parametrize("fdatabasis, expected_std_coefficients", [
```
vnmabus, line 137: Same as above, splitting in several tests is more readable, it is clearer when one test fail, and the duplication is almost inexistent. Parameterized fixtures are better for when the test is the same but with a different parameter (e.g., test that something works with both `FourierBasis` and `BSplineBasis`).

vnmabus (changes requested):

skfda/exploratory/stats/_stats.py
```diff
@@ -99,6 +100,63 @@ def cov(
     return X.cov(ddof=ddof)
 
 
+@functools.singledispatch
+def std(X: F, ddof: int = 1) -> F:
```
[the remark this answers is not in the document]
pcuestas: Yes, I think we did. `ddof=1` is the default in `numpy.cov`. However, I can change `var` and `std` so that they have the same defult `ddof` values as `numpy.var` and `numpy.std`: both `ddof=0`.

skfda/exploratory/stats/_stats.py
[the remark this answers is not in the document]
pcuestas: What I meant is that if `sample_names=None`, then `X.copy().sample_names` defaults to `X.sample_names`, which has more names than there are samples in `std(X)`. This causes the copy to fail.

skfda/exploratory/stats/_stats.py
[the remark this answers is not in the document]
pcuestas: Yes, I think we did. `ddof=1` is the default in `numpy.cov`. However, I can change `var` and `std` so that they have the same default `ddof` values as `numpy.var` and `numpy.std`: both `ddof=0`.

skfda/exploratory/stats/_stats.py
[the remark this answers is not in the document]
vnmabus: Oh, I see, that is an unfortunate value for the sentinel. Maybe we should change that in the future. However, you can still pass a tuple with one `None` value.

skfda/exploratory/stats/_stats.py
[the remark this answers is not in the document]
vnmabus: I thought that we decided the opposite (`ddof = 0`). Please bring this up in the next meeting for clarification. In the future please add the decisions and rationale as comments in the original issue or PR, for future reference.

skfda/_utils/_utils.py
```diff
@@ -605,3 +604,36 @@ def _classifier_get_classes(
...
+
+    Returns:
+        FDataBasis: FDataBasis with calculated coefficients and the new
+        basis.
+    """
+    from .. import FDataBasis
+    from ..misc._math import inner_product_matrix
+
+    if isinstance(f, FDataBasis) and f.basis == new_basis:
+        return f
```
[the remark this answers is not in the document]
pcuestas: I think returning a copy is safer. Should it be a deep copy?

skfda/_utils/_utils.py
```diff
@@ -605,3 +604,36 @@ def _classifier_get_classes(
...
+
+    Args:
+        f: math function.
+        new_basis: the basis of the output.
+
+    Returns:
+        FDataBasis: FDataBasis with calculated coefficients and the new
+        basis.
+    """
+    from .. import FDataBasis
```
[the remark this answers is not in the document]
pcuestas: A NameError is thrown if removed. Should FDataBasis be included on top of the file?

```bash
FAILED skfda/tests/test_stats_std.py::test_std_fdatabasis_vector_valued_basis[FourierBasis-5-3] - NameError: name 'FDataBasis' is not defined
```

skfda/exploratory/stats/_stats.py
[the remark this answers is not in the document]
pcuestas: Done.

skfda/_utils/_utils.py
```diff
@@ -605,3 +604,36 @@ def _classifier_get_classes(
...
+    from ..misc._math import inner_product_matrix
+
+    if isinstance(f, FDataBasis) and f.basis == new_basis:
+        return f
+
+    inner_prod = inner_product_matrix(
```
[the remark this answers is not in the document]
pcuestas: Don't `inner_product` and `inner_product_matrix` accept callables already?

https://github.com/GAA-UAM/scikit-fda/blob/0457e39e02119f1bfa675e422d68839aff2fda1d/skfda/misc/_math.py#L214-L221
https://github.com/GAA-UAM/scikit-fda/blob/0457e39e02119f1bfa675e422d68839aff2fda1d/skfda/misc/_math.py#L506-L510

And 

https://github.com/GAA-UAM/scikit-fda/blob/0457e39e02119f1bfa675e422d68839aff2fda1d/skfda/misc/_math.py#L22-L25

skfda/tests/test_stats_std.py
pcuestas: I think it was not well designed and that the newer tests are clearer. I removed this one.

skfda/exploratory/stats/_stats.py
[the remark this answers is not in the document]
pcuestas: Understood.

skfda/exploratory/stats/_stats.py
```diff
@@ -99,6 +100,64 @@ def cov(
...
+    basis = X.basis
+    coeff_matrix = np.cov(X.coefficients, rowvar=False, ddof=ddof)
+
+    def std_function(t_points: NDArrayFloat) -> NDArrayFloat:
+        assert len(t_points) == 1, (
+            "Standard deviation function only implemented for "
+            "one-point-at-a-time evaluations."
+        )
+        basis_evaluation = basis(t_points).reshape((-1, 1))
+        return np.sqrt(
```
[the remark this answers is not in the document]
pcuestas: It was not on our side. `np.sqrt`'s return type was not recognized as ndarray, but I managed to fix it by calling `np.reshape()` instead of `.reshape()`. Perhaps this sidestepping is questionable.

skfda/exploratory/stats/_stats.py
```diff
@@ -99,6 +100,66 @@ def cov(
...
+            basis_evaluation.T @ coeff_matrix @ basis_evaluation,
+        ).reshape((1, -1, 1))
+
+    return function_to_fdatabasis(f=std_function, new_basis=X.basis)
```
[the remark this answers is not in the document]
pcuestas: I believe this is related to https://github.com/GAA-UAM/scikit-fda/pull/566#discussion_r1328077971 and not being able to import FDataBasis in _utils.py due to circular import issues.

skfda/_utils/_utils.py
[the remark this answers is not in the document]
vnmabus: Sorry, my fault. Including it at the top would cause a circular reference. Just silence the warning with an apropriate comment.

skfda/_utils/_utils.py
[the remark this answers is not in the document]
vnmabus: They accept them at runtime. However, they are typed as accepting `Vector`, a protocol which defines sum and scalar product. Normal callables do not define these operations (`FData` are an exception).

skfda/exploratory/stats/_stats.py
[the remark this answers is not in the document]
vnmabus: It is ok for me. Although I do not get why the `reshape` method has different typing.

skfda/_utils/_utils.py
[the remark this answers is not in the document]
pcuestas: OK, thanks.

skfda/exploratory/stats/_stats.py
```diff
@@ -99,6 +100,59 @@ def cov(
...
+
+    Returns:
+        Standard deviation of all the samples in the original object, as a
+        :term:`functional data object` with just one sample.
+
+    """
+    raise NotImplementedError("Not implemented for this type")
+
+
+@std.register(FDataGrid)
```
vnmabus, line 127: Is it really necessary to specify the type here?

skfda/exploratory/stats/_stats.py
```diff
@@ -99,6 +100,59 @@ def cov(
...
+
+    basis = X.basis
+    coeff_cov_matrix = np.cov(
+        X.coefficients, rowvar=False, ddof=ddof,
+    ).reshape((basis.n_basis, basis.n_basis))
+
+    def std_function(t_points: NDArrayFloat) -> NDArrayFloat:  # noqa: WPS430
+        basis_evaluation = basis(t_points).reshape((basis.n_basis, -1))
+        std_values = np.sqrt(
+            np.diag(basis_evaluation.T @ coeff_cov_matrix @ basis_evaluation),
```
vnmabus, line 149: Is there a better way that does not need to compute the whole matrix just to keep the diagonal?

skfda/tests/test_stats_std.py
```diff
@@ -0,0 +1,209 @@
...
+# Fixtures for test_std_fdatabasis_vector_valued_basis
+
+
+@pytest.fixture(params=[3, 5])
+def vv_n_basis1(request: Any) -> int:
+    """n_basis for 1st coordinate of vector valued basis."""
+    return request.param  # type: ignore[no-any-return]
+
+
+@pytest.fixture(params=[3])
```
vnmabus, line 30: There is only one value. Why parameterize it then?

skfda/tests/test_stats_std.py
```diff
@@ -0,0 +1,209 @@
...
+def vv_n_basis2(request: Any) -> int:
+    """n_basis for 2nd coordinate of vector valued basis."""
+    return request.param  # type: ignore[no-any-return]
+
+
+@pytest.fixture
+def vv_basis1(vv_n_basis1: int) -> Basis:
+    """1-dimensional basis to test for vector valued basis."""
+    # First element of the basis is assumed to be the 1 function
```
vnmabus, line 39: WDYM? That is not true for the B-spline basis.

skfda/tests/test_stats_std.py
```diff
@@ -0,0 +1,209 @@
...
+    return request.param(  # type: ignore[no-any-return]
+        domain_range=(0, 1), n_basis=vv_n_basis2,
+    )
+
+
+# Fixtures for test_std_fdatabasis_tensor_basis
+
+@pytest.fixture(params=[3])
```
vnmabus, line 58: Again, there is no need for parametrization in this and next fixtures.

skfda/exploratory/stats/_stats.py
pcuestas: It is not. Removed.

skfda/exploratory/stats/_stats.py
pcuestas: There are three alternatives that I can think of:

* diag:
```python
        std_values = np.sqrt(
            np.diag(basis_evaluation.T @ coeff_cov_matrix @ basis_evaluation),
        )
```
* einsum:
```python
        std_values = np.sqrt(
            np.einsum("bi,bd,di->i", basis_evaluation, coeff_cov_matrix, basis_evaluation),
        )
```

* sum:
```python
        std_values = np.sqrt(np.sum(
            basis_evaluation * (coeff_cov_matrix @ basis_evaluation), axis=0,
        ))
```

After doing some quick profiling, I cannot see any difference between the three. I think the 'diag' version is the clearest by being the most explicit.

skfda/tests/test_stats_std.py
pcuestas: I should not have used BSpline basis. Fixed.

skfda/exploratory/stats/_stats.py
eliegoudout: Hello,

I *pre-post* from my phone in order to speak before a potential a merge :)

I just wanted to express a small disagreement with @pcuestas.

> After doing some quick profiling, I cannot see any difference between the three in terms of time.

I personally found a consistent 60~80% performance difference (`diag_time / sum_time`) on my single cpu for `NxN` matrices with `N > 64`.

Before showing the results yesterday, I wanted to render nicer graphs with more samples, uncertainty, etc… and ended up not having the time to post. And I probably won’t have enough time before another 5 days now (hence the teaser comment 🤣).

I will definitely come back to you with a proper visual in a few days!

One thing to note: if `A, B, C` are C-contiguous matrices, then

```
np.sum(A @ B * C.T, axis=-1)
```

is 10%~20% slower than

```
np.sum(A * C.T @ B.T, axis=-1)
```

For the same output. I included this gain in the first mentioned 60%~80% overall gain.

I'll be back :)
Cheers!

pcuestas: My "quick profiling" was definitely not specific enough. I was measuring the performance of the std_fdatabasis function and not the actual numpy operations. I'll share more accurate measurements when I have them ready. 

The sum option seems to be faster than diag. Einsum is the slowest by a factor of ~10.

pcuestas: I have compared the execution time of the following functions (which output the same result):

```python
# b is ndarray of shape (B, n_points)
# C is ndarray of shape (B, B)
# m_result is ndarray of shape (n_points,)

def m_diag(b, C):
    return np.diag(b.T @ C @ b)


def m_sum(b, C):
    return np.sum(b * (C @ b), axis=0)


def m_sum_T(b, C):
    return np.sum(b.T * (b.T @ C.T), axis=1)


def m_einsum(b, C):
    return np.einsum('bi,bd,di->i', b, C, b)
```

by doing:

```python
def time_func(func, *args, n_times):
    start = time.time()
    for _ in range(n_times):
        func(*args)
    end = time.time()
    return end - start
```

Full code I have written for this is in https://gist.github.com/pcuestas/46be0f88aa4a14d481845cc5d646c8e0

Times and speedups that I got (time data also present in previously linked gist):

Figure_1

It seems that for big enough matrix sizes the speedup of `sum` over `diag` stabilizes around 1.5 (`sum` is 50% faster than `diag`) on my machine.

I am not sure whether the procedure I followed was good enough, I do not have much experience with this.

eliegoudout: Hello, I can see thaty you profiled with `time.time`, which is generally considered bad practise for profiling. People tend to use `timeit` or `time.perf_counter` directly (which is used by `timeit` by default), for more precision. Also, it would be beneficial to plot the results on a logscale to interpret them more easily I think.

For my part, testing these:
```
def diag(A, B, C):
    return np.diag(A @ B @ C)

def sum1(A, B, C):
    return np.sum(A @ B * C.T, axis=-1)

def sum2(A, B, C):
    return np.sum(A * (C.T @ B.T), axis=-1)
```

over multiple runs (cf graph), I got the following results with barely noticeable uncertainty (Q1 ↔ Q3) areas:
profiling

We can observe that depending on the size of the matrices, it is not always obvious what will happen. But what is sure, is that indeed, the C-contiguousity can be leveraged, no matter the size of the matrices.

vnmabus: You should probably report the minimum instead of median, according to the following comment in [`timeit.repeat`](https://docs.python.org/3/library/timeit.html#timeit.Timer.repeat):
> It’s tempting to calculate mean and standard deviation from the result vector and report these. However, this is not very useful. In a typical case, the lowest value gives a lower bound for how fast your machine can run the given code snippet; higher values in the result vector are typically not caused by variability in Python’s speed, but by other processes interfering with your timing accuracy. So the [min()](https://docs.python.org/3/library/functions.html#min) of the result is probably the only number you should be interested in. After that, you should look at the entire vector and apply common sense rather than statistics.

It probably won't change much, though.

As for the methods compared, another approach would be to compute the Cholesky decomposition of $B = LL^T$ outside the inner function, given that it is a covariance matrix. Then we could do $C = L^T A$ and finally compute the sum of the rows of the elementwise square $C^2$ (the squared norm of C rows). This uses the fact that $A^TBA = C^T C$ and the fact that the diagonal of that is the vector of squared norms.

pcuestas: The suggested computation by precalculating the Cholesky decomposition has basically the same performance as the `sum` version. The reason why this happens is that both require the same operations (once the Cholesky decomposition has been precomputed):

```python

def m_sum(b, C):
    return np.sum(b * (C @ b), axis=0)

def m_cholesky(b, L_T):
    """C = L @ L.T and L_T = L.T"""
    return np.sum((L_T @ b)**2, axis=0)
```

As the `m_cholesky` version is less explicit (and requires the Cholesky decomposition of the coefficient covariances matrix), it was decided on the meeting from 2023-10-10 to use the `sum` version:

```python
std_values = np.sqrt(np.sum(
    basis_evaluation * (coeff_cov_matrix @ basis_evaluation), axis=0,
))
```

vnmabus (approved):
LGTM.

PR #578, opened 2023-10-04
Change `ddof` argument to `correction` in var and cov functions

### LINKED ISSUE #576: Change ddof argument to correction in var and cov
Modify the `var` and `cov` functions so that the `ddof` (Delta Degrees of Freedom) parameter is renamed to `correction`, to match the Array API Standard: https://data-apis.org/array-api/latest/API_specification/generated/array_api.var.html

pcuestas:
Change `ddof` argument name to `correction` to match the Python array API standard.

Closes #576

### DIFF
--- a/skfda/exploratory/stats/_stats.py
+++ b/skfda/exploratory/stats/_stats.py
@@ -41,22 +41,22 @@ def mean(
     return (X * weight).sum()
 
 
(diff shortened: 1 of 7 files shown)


vnmabus (approved):
LGTM

PR #582, opened 2023-10-12
Set `correction=0` by default

### LINKED ISSUE #581: Default correction=0 for cov, var
Change the default value of the `correction` parameter in the var and cov functions to `0` to adhere to the [Python array API standard](https://data-apis.org/array-api/latest/API_specification/generated/array_api.var.html).

pcuestas:
(and explicitly set the correction value where it was not set, to avoid changing the previous behavior of the modules).

Closes #581


vnmabus (approved):
LGTM

PR #608, opened 2024-04-01
Integrate `FDataIrregular` using `scipy.integrate.simpson` on each curve separately 

pcuestas:

Integration for irregular curves has been implemented using `scipy.integrate.simpson` to integrate irregular curves (of 1-dimensional domain) one by one.

## Additional information

We have not agreed on how integration of irregular data with higher dimensional domains should be defined, so it is not to be implemented yet.

This is a first implementation which may or may not be later improved by manually computing simpson's coefficients, multiply them against the curves' values and calling `skfda.representation.irregular._reduceat` to add the results.

### DIFF
--- a/skfda/representation/irregular.py
+++ b/skfda/representation/irregular.py
@@ -11,7 +11,6 @@ import itertools
 import numbers
 from typing import (
     Any,
-    Callable,
     Optional,
     Sequence,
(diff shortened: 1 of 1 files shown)


vnmabus (approved):
LGTM.

PR #612, opened 2024-04-11
Euler_maruyama function

psotom:
Included a function that makes numerical integration of an Îto SDE using the euler_maruyama shceme. It has been included in the datasets module to extend the functionality dedicated to generate synthetic data.

- [x] I have performed a self-review of my code
- [ ] The code conforms to the style used in this package
- [x] The code is fully documented and typed (type-checked with [Mypy](https://mypy-lang.org/))
- [x] I have added thorough tests for the new/changed functionality

### DIFF
--- a/skfda/datasets/_samples_generators.py
+++ b/skfda/datasets/_samples_generators.py
@@ -47,7 +47,6 @@ def euler_maruyama(  # noqa: WPS210
     start: float = 0.0,  # noqa: WPS358 -- Distinguish float from integer
     stop: float = 1.0,
     diffusion_matricial_term: bool = True,
-    dim_noise: int | None = None,
     random_state: RandomStateLike = None,
 ) -> FDataGrid:
     r"""Numerical integration of an Itô SDE using the Euler-Maruyana scheme.
@@ -222,11 +221,11 @@ def euler_maruyama(  # noqa: WPS210
             noise,
         )
 
-    if dim_noise is None:
-        dim_noise = dim_codomain
+    dim_noise = dim_codomain
 
     if diffusion_matricial_term:
         diffusion_times_noise = matrix_diffusion_times_noise
+        dim_noise = diffusion_function(start, initial_values).shape[-1]
     else:
         diffusion_times_noise = vector_diffusion_times_noise
 
--- a/skfda/datasets/_samples_generators.py
+++ b/skfda/datasets/_samples_generators.py
@@ -20,56 +20,60 @@ MeanCallable = Callable[[np.ndarray], np.ndarray]
 CovarianceCallable = Callable[[np.ndarray, np.ndarray], np.ndarray]
 
 MeanLike = Union[float, NDArrayFloat, MeanCallable]
+SDETerm = Callable[[float, NDArrayFloat], NDArrayFloat]
 
 
 class InitialValueGenerator(Protocol):
-    """Class to represent SDE initial value generators."""
+    """Class to represent SDE initial value generators.
+
+    This is intented to be an interface compatible with the rvs method of
+    SciPy distributions.
+    """
 
     def __call__(
         self,
         size: int,
         random_state: RandomStateLike,
-    ) -> np.typing.NDArray[np.floating[Any]]:
+    ) -> NDArrayFloat:
         """Interface of initial value generator."""
-        ...
 
 
 def euler_maruyama(
     initial_condition: ArrayLike | InitialValueGenerator,
     n_grid_points: int = 100,
-    drift: Callable[[float, np.typing.NDArray[np.floating[Any]]],
-                    np.typing.NDArray[np.floating[Any]]] | None = None,
-    diffusion: Callable[[float, np.typing.NDArray[np.floating[Any]]],
-                        np.typing.NDArray[np.floating[Any]]] | None = None,
+    drift: SDETerm | ArrayLike = 0,
+    diffusion: SDETerm | ArrayLike = 1,
     n_samples: int | None = None,
     start: float = 0,
     stop: float = 1.0,
+    is_diffusion_matrix: bool = False,
     dim_noise: int | None = None,
     random_state: RandomStateLike = None,
 ) -> FDataGrid:
-    """Numerical integration of an Itô SDE using the Euler-Maruyana scheme.
+    r"""Numerical integration of an Itô SDE using the Euler-Maruyana scheme.
 
     .. math::
 
-        d X(t) = f(t, X(t)) d t + g(t,X(t)) d W(t).
+        d X(t) = f(t, X(t)) d t + g(t, X(t)) d W(t).
 
     Args:
-        initial_condition: Initial condition of the SDE. It can have one of
+        initial_condition: initial condition of the SDE. It can have one of
             three formats: An starting initial value from which to
             calculate *n_samples* trajectories. An array of initial values.
             For each starting point a trajectory will be calculated. A
             function that generates random numbers or vectors. It should
             have two parameters called size and random_state and it should
             return an array.
-        n_grid_points: The total number of points of evaluation.
+        n_grid_points: the total number of points of evaluation.
         drift: drift coefficient (:math:`f(t,X_t)` in the equation).
         diffusion: diffusion coefficient (:math:`g(t,X_t)` in the equation).
-        n_samples: Number of trajectories integrated.
-        start: Starting time of the trajectories.
-        stop: Ending time of the trajectories.
+        n_samples: number of trajectories integrated.
+        start: starting time of the trajectories.
+        stop: ending time of the trajectories.
+        is_diffusion_matrix: True if the diffusion coefficient is a matrix.
         dim_noise: dimension of the noise factor. By default is the data
             dimension.
-        random_state: Random state.
+        random_state: random state.
 
 
     Returns:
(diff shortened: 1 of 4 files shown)


skfda/datasets/_samples_generators.py
```diff
@@ -1,26 +1,212 @@
...
+    if diffusion is None:
+        diffusion = default_diffusion
+
+    test_diffusion_dim = np.atleast_1d(diffusion(start, initial_values))
+
+    if test_diffusion_dim.ndim == 1:
+        if len(test_diffusion_dim) not in {dim_codomain, 1}:
+            raise ValueError(
+                f"The diffusion function returns the wrong dimensions."
+                f"The expected dimension is {dim_codomain} or 1.",
```
vnmabus, line 163: Why 1?

skfda/datasets/_samples_generators.py
```diff
@@ -481,7 +667,7 @@ def make_random_warping(
     np.square(v, out=v)
 
     # Creation of FDataGrid in the corresponding domain
-    data_matrix = scipy.integrate.cumulative_trapezoid(
+    data_matrix = scipy.integrate.cumtrapz(
```
vnmabus, line 670: Why did you change the name here? `cumulative_trapezoid` is the modern name, and `cumtrapz` is deprecated.

skfda/tests/test_euler_maruyama.py
```diff
@@ -0,0 +1,711 @@
...
+
+
+def test_one_initial_point() -> None:
+    """Case 1 -> One initial point + n_samples > 0.
+
+    Tests:
+        - initial_condition = 0, n_samples = 2
+        - initial_condition = [0], n_samples = 2.
+            # The result should be the same as the first one
+        - initial_condition = np.array([0, 0]), n_samples = 2.
```
vnmabus, line 18: Shouldn't the last one be a separate test?

skfda/tests/test_euler_maruyama.py
```diff
@@ -0,0 +1,711 @@
...
+
+    """
+    n_samples = 2
+    n_grid_points = 5
+    random_state = np.random.RandomState(1)
+    # Case 1 Starting point = float + n_samples
+    initial_float = 0
+
+    expected_result = np.array([[
```
vnmabus, line 27: How are the expected results determined?

vnmabus (changes requested):

skfda/datasets/_samples_generators.py
psotom: The value 1 references to the case were the diffusion coefficient is just a constant. In that case, test_diffusion_dim should be an array with one element and that would be a valid scenario.

skfda/datasets/_samples_generators.py
```diff
@@ -1,26 +1,212 @@
...
+            function that generates random numbers or vectors. It should
+            have two parameters called size and random_state and it should
+            return an array.
+        n_grid_points: The total number of points of evaluation.
+        drift: drift coefficient (:math:`f(t,X_t)` in the equation).
+        diffusion: diffusion coefficient (:math:`g(t,X_t)` in the equation).
+        n_samples: Number of trajectories integrated.
+        start: Starting time of the trajectories.
+        stop: Ending time of the trajectories.
+        dim_noise: dimension of the noise factor. By default is the data
```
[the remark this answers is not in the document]
psotom: Sometimes, when a multidimensional process is generated, the sources of noise are less than the number of dimensions of the process. In that case, the difussion is represented by a non-square matrix, which is multiplied by the noise vector at each iteration. We need this parameter to distinguish this case from the normal one.

skfda/tests/test_euler_maruyama.py
psotom: The euler_maruyama function generates random noise to calculate the trajectories of the solutions of the SDE. In order to calculate the expected result, data is generated using np.random.RandomState and knowing the initial_condition, drift and diffusion the expected result is manually calculated.

skfda/datasets/_samples_generators.py
```diff
@@ -1,24 +1,217 @@
...
+        return np.einsum(
+            '...dj, ...j -> ...d',
+            formatted_diffusion(t_n, x_n),
+            noise,
+        )
+
+    if dim_noise is None:
+        dim_noise = dim_codomain
+
+    if is_diffusion_matrix:
```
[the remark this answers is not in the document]
psotom: When the user does not specify the dimension of the noise, we suppose that is the same as the data dimension as this is the standard case.

skfda/datasets/_samples_generators.py
```diff
@@ -1,24 +1,217 @@
...
+        diffusion_times_noise = matrix_diffusion_times_noise
+    else:
+        diffusion_times_noise = vector_diffusion_times_noise
+
+    data_matrix = np.zeros((n_samples, n_grid_points, dim_codomain))
+    times = np.linspace(start, stop, n_grid_points)
+    delta_t = times[1:] - times[:-1]
+    noise = random_state.standard_normal(
```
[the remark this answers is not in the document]
psotom: The noise models an instance of a [Wiener process](https://en.wikipedia.org/wiki/Wiener_process). Increments of the Wiener process follow a normal distribution with variance equal to the time difference (it does not depend on other parameters). If you want to create a process with a higher standard deviation / volatily, you can achieve it simply by making the diffusion term bigger.

vnmabus:
So, for me the only pending issues for accepting this PR are the parameters `diffusion_matricial_term` and `dim_noise`:

- For `diffusion_matricial_term`, I do not get what does that mean in the model. Please, explain it to me (with equations, if possible) to see if it is necessary.
- From the equation it seems to me that the parameter `dim_noise` can be computed from the `diffusion`, and should not be passed by the user. Again, if that is not correct, please explain.

psotom:
> So, for me the only pending issues for accepting this PR are the parameters `diffusion_matricial_term` and `dim_noise`:
> 
> * For `diffusion_matricial_term`, I do not get what does that mean in the model. Please, explain it to me (with equations, if possible) to see if it is necessary.
> * From the equation it seems to me that the parameter `dim_noise` can be computed from the `diffusion`, and should not be passed by the user. Again, if that is not correct, please explain.

To answer your question, I will explain some context:

The formula of the SDE we want to model is the following:

$$d\mathbf{X}(t) = \mathbf{F}(t, \mathbf{X}(t))dt + \mathbf{G}(t, \mathbf{X}(t))d\mathbf{W}(t).$$

In this equation, 

- $\mathbf{X} = (X^{(1)}, X^{(2)}, ... , X^{(q)})\in \mathbb{R}^q$ is a vector that represents the stochastic process whose values we wish to simulate. The superindeces indicate the coordinate of the vector.
 - The function $\mathbf{F}: \mathbb{R} \times \mathbb{R}^q \rightarrow \mathbb{R}^q$ defines the drift term, which is the deterministic component of the equation.
 - The function $\mathbf{G}: \mathbb{R} \times \mathbb{R}^q \rightarrow \mathbb{R}^q \times \mathbb{R}^m$ defines the diffusion term and refers to the stochastic component of the evolution. 
- $\mathbf{W}(t) \in \mathbb{R}^m$ refers to a Wiener process (Standard Brownian motion) of dimension $m$. The components of this process are independent.
- $q$ refers to the dimension of the variable $\mathbf{X}$ (dimension of the codomain).
- $m$ is the dimension of the noise.
	
-------------
diffusion_matricial_term justification:

The main reason for the use of this parameter is to add flexibility to the input of the SDE arguments. 

As explained above the diffusion term is a matrix. However, many times is independent on $X$ and we can consider it as a scalar. We have two options:

- Treat everything as a matrix. For the 1-dimensional case an input such as diffusion=1 is not valid and diffusion=[[1]] would be needed. Also, if $\mathbf{X}$ is multidimensional, if you want to add a constant diffusion, it must given by a multiple of the identity matrix with the corresponding dimensions.
- Distinguish between scalar and matrix diffusion terms. In this case, we can define for example diffusion=1 and it will work no matter what are the dimensions of the process. In general, this approach facilitates a lot the way the diffusion can be defined. For this distinction we cannot only rely on the diffusion function, but we need the diffusion_matricial_term parameter.

Additionally, if we choose option 2 we can add "for free" the possibility of the diffusion term being a vector. This is specially useful when we want to simulate a process in each of the coordinates.

For example, if I want to simulate a Geometric Brownian motion ($dX = \mu X dt + \sigma X dW_t$) in each coordinate, the definition of the diffusion for both options would be:

```python
def gbm_option1_diffusion(t: float, x: NDArrayFloat) -> NDArrayFloat:
    """I must define a multiple of the identity matrix for this case."""
    diffusion = np.zeros((n_samples, dim_codomain, dim_codomain))
    for i in np.arange(dim_codomain):
        diffusion[:, i, i] = sigma * x[:, i]
    return diffusion

def gbm_option2_diffusion(t: float, x: NDArrayFloat) -> NDArrayFloat:
    return sigma * x
````

Wrapping everything up, I think option 2 (which includes diffusion_matricial_term) is better because it increases a lot the flexibility in the input of the function. 

----------------------
dim_noise:

After close examination I think I have found a way to compute the dimension of the noise without needing the parameter. I will make a commit.

vnmabus:
But what I do not get is why it is not possible to just detect the appropriate version depending on the `diffusion` parameter passed, instead of forcing the user to pass an explicit boolean parameter. Is there any situation that could lead to ambiguity?

vnmabus:
I think I see the problem better now. I will accept this as is, and lets see if it is possible to remove it in the future.

vnmabus (approved):
LGTM

vnmabus:
@all-contributors Please add @psotom for code, example, ideas, test, research and documentation.

PR #672, opened 2025-04-23
RUFF format correction Lp metrics

luisheb:

Corrected the style for the Lp norm and distance files in accordance to the new RUFF specifications 

### DIFF
--- a/skfda/misc/metrics/_lp_distances.py
+++ b/skfda/misc/metrics/_lp_distances.py
@@ -1,5 +1,5 @@
-
 """Implementation of Lp distances."""
+
 from __future__ import annotations
 
 import math
@@ -18,7 +18,7 @@ if TYPE_CHECKING:
 T = TypeVar("T", NDArrayFloat, FData)
 
 
-class LpDistance(NormInducedMetric[NDArrayFloat| FData]):
+class LpDistance(NormInducedMetric[NDArrayFloat | FData]):
     r"""
     Lp distance for functional data objects.
 
--- a/skfda/misc/metrics/_lp_distances.py
+++ b/skfda/misc/metrics/_lp_distances.py
@@ -3,21 +3,22 @@
 from __future__ import annotations
 
 import math
-from typing import Optional, TypeVar, Union
+from typing import TYPE_CHECKING, Final, TypeVar
 
 import numpy as np
-from typing_extensions import Final
 
 from ...representation import FData
-from ...typing._metric import Norm
 from ...typing._numpy import NDArrayFloat
 from ._lp_norms import LpNorm
 from ._utils import NormInducedMetric, pairwise_metric_optimization
 
+if TYPE_CHECKING:
+    from ...typing._metric import Norm
+
 T = TypeVar("T", NDArrayFloat, FData)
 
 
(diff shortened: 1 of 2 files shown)


skfda/misc/metrics/_lp_distances.py
```diff
@@ -3,21 +3,22 @@
...
 from ._utils import NormInducedMetric, pairwise_metric_optimization
 
+if TYPE_CHECKING:
+    from ...typing._metric import Norm
+
 T = TypeVar("T", NDArrayFloat, FData)
 
 
-class LpDistance(NormInducedMetric[Union[NDArrayFloat, FData]]):
+class LpDistance(NormInducedMetric[NDArrayFloat| FData]):
```
vnmabus: I think a space is missing here before the pipe operator. I am actually surprised this is not flagged by Ruff or WPS.

skfda/misc/metrics/_lp_norms.py
```diff
@@ -89,20 +86,26 @@ class LpNorm:
...
-        return f"{type(self).__name__}(" f"p={self.p}, vector_norm={self.vector_norm})"
-
-    def __call__(self, vector: Union[NDArrayFloat, FData]) -> NDArrayFloat:
+        return (
+            f"{type(self).__name__}("
+            f"p={self.p}, vector_norm={self.vector_norm})"
+        )
+
+    def __call__(  # noqa: C901
+        self, vector: NDArrayFloat | FData,
```
vnmabus: Put each parameter in its own line, or all the signature in the same line, nothing in between. It is a shame because this was previously flagged by WPS, but as they removed the extension that did this, and it is not yet implemented in Ruff, it is not flagging this anymore.

skfda/misc/metrics/_lp_norms.py
```diff
@@ -131,7 +134,7 @@ def integrand(*args: NDArrayFloat) -> NDArrayFloat:  # noqa: WPS430
 
                 try:
                     f1 = call(f_args)[:, 0, :]
-                except Exception:
+                except Exception:  # noqa: BLE001
```
vnmabus: Actually Ruff is on the right here. We should store `call(f_args)` in a variable, instead of computing this twice in the worst case, and catch the index error afterwards.

vnmabus (changes requested):

skfda/misc/metrics/_lp_norms.py
eliegoudout: I just tried locally and ruff doees detect this on my side
image

eliegoudout: To be clear, `ruff format` not `ruff check`.

vnmabus: So, Ruff check does not flag anything, but Ruff format changes that anyway? That seems weird.

skfda/misc/metrics/_lp_norms.py
luisheb: Ruff flagged it because of the blind except, but the proper fix is definitely to avoid recomputing call(f_args). I've updated it accordingly and corrected both issues. Thanks for pointing it out!

vnmabus (approved):
LGTM

PR #673, opened 2025-04-23
MixedDataPlot Class

luisheb:

This PR introduces a new MixedDataPlot class for visualizing pandas DataFrames that contain a combination of functional (FData) and scalar (numerical or array-like) data. The class inherits from BasePlot and extends the plotting capabilities of scikit-fda to support mixed data structures in a structured grid layout.


luisheb:
What’s the recommended approach for testing functionality that generates plots? Should we rely on checking object properties (like axes titles, number of lines, etc.), or is there a preferred testing strategy already in place for visualizations?
@vnmabus

PR #676, opened 2025-04-24
Patch for RUFF linting

luisheb:
All existing RUFF errors have been ignored to ensure that future reports focus only on new changes. Addressing the existing errors will be handled separately as a pending issue.


luisheb:
@vnmabus  This is a proposal to prevent existing RUFF errors from interfering with new changes. Let me know your thoughts.

vnmabus:
I would rather not do this. It changes a lot of files at the same time, creating a lot of noise in Git blame and potential conflicts, just as a temporary "solution". Moreover, it won't be clear in the future which errors were silenced on purpose and which ones were not. 

What I propose instead is to adopt the solution suggested in the commit message of https://github.com/astral-sh/ruff-action/commit/f2e32211077f40bfeab40e16285216924ebea4cb, so that the Ruff action runs only on the files that have been changed. What do you think about that?

luisheb:
You're right, I agree. I wasn’t aware that was an option, and it’s definitely a much cleaner solution.

vnmabus:
> You're right, I agree. I wasn’t aware that was an option, and it’s definitely a much cleaner solution.

Do you want to try to do the PR yourself?

luisheb:
Sure! I'll start working on it and let you know if I run into any issues.

PR #683, opened 2025-05-09
Fix/cast error f data basis

luisheb:

One line change in FDataBasis equals function


vnmabus (approved):
LGTM.

PR #687, opened 2025-05-12
Mixed data visualization

luisheb:

Functionality to plot Mixed data objects represented as pandas `DataFrame`. 


examples/representation/plot_mixed_data.py
```diff
@@ -0,0 +1,149 @@
...
+separately as independent entries. We also explore how to structure and
+visualize this data using `pandas` and scikit-fda's visualization tools.
+"""
+
+# Author: Luis Hebrero Garicano
+# License: MIT
+# sphinx_gallery_thumbnail_number = 1
+
+
+# %% [markdown]
```
vnmabus: Why this `[markdown]` annotation? Does it do anything?

examples/representation/plot_mixed_data.py
```diff
@@ -0,0 +1,149 @@
...
+fd_vector.plot()
+
+# %% [markdown]
+# We now create a mixed data object using a `pandas.DataFrame`. This includes:
+# - a scalar variable: the climate zone,
+# - functional variables: the temperature and its derivative,
+# - and the vector-valued version combining both.
+#
+# This illustrates two valid ways to include a function and its derivative
+# in mix
```
vnmabus:
```suggestion
# as part of the same observation.
```

examples/representation/plot_mixed_data.py
```diff
@@ -0,0 +1,149 @@
...
+    {
+        "category": y,
+        "temperatures": fd_temperatures,
+        "temperature_derivatives": fd_1st_temperatures_smooth,
+        "precipitations_vector": fd_vector,
+    },
+)
+
+# %% [markdown]
+# Finally, we use `plot_mixed_data` to visualize the full mixed dataset. Each
```
vnmabus: Add a reference.

luisheb:
I'm not entirely satisfied with how the graphs look in the final [example](https://fda--687.org.readthedocs.build/en/687/auto_examples/representation/plot_mixed_data.html#sphx-glr-auto-examples-representation-plot-mixed-data-py), but the core plotting logic is complete and ready for review.

skfda/exploratory/visualization/representation.py
```diff
@@ -94,23 +103,26 @@ def _get_color_info(
         if group_names_array is not None:
             patches = [
                 matplotlib.patches.Patch(color=c, label=l)
-                for c, l in zip(group_colors_array, group_names_array)
+                for c, l in zip(
+                    group_colors_array,
+                    group_names_array,
+                    strict=False,
```
vnmabus: Why `False`?

skfda/exploratory/visualization/representation.py
```diff
@@ -778,3 +780,339 @@ def set_color_dict(
...
+
+    Functional variables are plotted using their native `.plot()` method.
+    Vector valued functional data generate multiple subplots, one for each
+    component. Numerical data are shown as scatter plots.
+
+    Supports grouping samples by color using `group`, with optional custom
+    colors (`group_colors`) and labels (`group_names`) for legend display.
+
+    Args:
+        df: pd.DataFrame object that we want to plot.
```
vnmabus: I
```suggestion
        df: DataFrame object that we want to plot.
```

skfda/exploratory/visualization/representation.py
```diff
@@ -778,3 +780,339 @@ def set_color_dict(
...
+        chart: figure over
```
vnmabus: I think it is better to have all parameter descriptions start with uppercase. I know we have not been consistent with this in the past.

skfda/exploratory/visualization/representation.py
```diff
@@ -778,3 +780,339 @@ def set_color_dict(
...
+            with the graphs are plotted or axis over where the graphs are
+            plotted. If None and ax is also None, the figure is
```
vnmabus: Format arguments and code in monospace. E.g.:

```suggestion
            plotted. If ``None`` and ``ax`` is also ``None``, the figure is
```

examples/representation/plot_mixed_data.py
```diff
@@ -0,0 +1,172 @@
...
+
+fig, axes = plt.subplots(1, 2, figsize=(8, 4))
+
+fd_temperatures.plot(axes=axes[0])
+fd_precipitations.plot(axes=axes[1])
+fig.tight_layout()
+plt.show()
+
+# %% [markdown]
+# To enrich the data with dynamic information, we compute the first derivative
```
vnmabus: What do you mean with "dynamic" in this context?

examples/representation/plot_mixed_data.py
```diff
@@ -0,0 +1,172 @@
...
+# of both temperature and precipitation curves. These derivatives can capture
+# local variation patterns such as rising or falling trends.
+
+fd_1st_temperatures = fd_temperatures.derivative(order=1)
```
vnmabus: I think you can omit the order for taking the first derivative.

examples/representation/plot_mixed_data.py
```diff
@@ -0,0 +1,172 @@
...
+)
+
+fig, axes = plt.subplots(1, 2, figsize=(8, 3))
+
+fd_vector.plot(axes=axes)
+fig.tight_layout()
+plt.show()
+
+# %% [markdown]
+# We now create a mixed data object using a `pandas.DataFrame`. This includes:
```
vnmabus: Add a reference to `pandas.DataFrame` using Sphinx.

examples/representation/plot_mixed_data.py
```diff
@@ -0,0 +1,172 @@
...
+# - a scalar variable: the climate zone (weather type),
```
vnmabus, line 141: The enumeration is not being shown correctly in the generated docs.

examples/representation/plot_mixed_data.py
```diff
@@ -0,0 +1,172 @@
...
+# - functional variables: the temperature (T(t)) and its derivative (T'(t)),
+# - and the vector-valued version combining both (P_vec(t)=(P(t), P'(t))).
```
vnmabus: I think you mean here the precipitations. From the writing it looks like the temperatures are included twice.

vnmabus (changes requested):

skfda/exploratory/visualization/representation.py
luisheb: Otherwise you get `zip()` without an explicit `strict=` parameterRuff [B905](https://docs.astral.sh/ruff/rules/zip-without-explicit-strict). The auto-fix action from Ruff adds strict=False. I believe its the default option just made explicit

examples/representation/plot_mixed_data.py
luisheb: “dynamic” refers to how the data changes over time. Adding information about the evolution of the original weather variables.

examples/representation/plot_mixed_data.py
luisheb: It's added by default when you convert a `.ipynb` to a `.py ` script with vs code. I will remove it for consistency with the rest of the examples.

skfda/exploratory/visualization/representation.py
vnmabus: Yes, it is the default option because it keeps existing semantics. But you have to consider if the two iterables are supposed to have the same length, and use `True` in that case.

examples/representation/plot_mixed_data.py
vnmabus: I am not sure this is clear as it is written, given that dynamic has many other uses in programming.

luisheb:
Any reference on how to write plot tests? Should I even do it? @vnmabus

vnmabus:
> Any reference on how to write plot tests? Should I even do it? @vnmabus

We do not have any infrastructure for that at the moment, unfortunately. So they are not mandatory.

examples/representation/plot_mixed_data.py
```diff
@@ -0,0 +1,177 @@
...
+# License: MIT
+# sphinx_gallery_thumbnail_number = 1
+
+
+# %%
+# We load the Canadian weather dataset. This dataset includes daily
+# temperature and precipitation curves for 35 weather stations in Canada,
+# along with a scalar variable: the climate zone of each station.
+
+import matplotlib.pyplot as plt
```
vnmabus: Could you move the imports to the first cell where they are needed? We are trying to do that in examples, as I feel it is less confusing to users.

examples/representation/plot_mixed_data.py
```diff
@@ -0,0 +1,177 @@
...
+import numpy as np
+import pandas as pd  # type: ignore[import-untyped]
```
vnmabus: Why the type ignore? I think Pandas provides type stubs.

vnmabus (changes requested):

eliegoudout:
> > Any reference on how to write plot tests? Should I even do it? @vnmabus
> 
> We do not have any infrastructure for that at the moment, unfortunately. So they are not mandatory.

My company will soon release a work project of mine, in which I developed an easy plot testing fixture. Would you be interested that I ping you here upon effective opensourcing?
I implemented a `match_plots` fixture which captures all calls to `plt.show()` and compares with expectations already saved. Updatig expectations occurs with `pytest --update-plots`.

vnmabus:
> My company will soon release a work project of mine, in which I developed an easy plot testing fixture. Would you be interested that I ping you here upon effective opensourcing? I implemented a `match_plots` fixture which captures all calls to `plt.show()` and compares with expectations already saved. Updatig expectations occurs with `pytest --update-plots`.

I planned to have a look to the tools that Matplotlib itself uses for testing. For example [`image_comparison`](https://matplotlib.org/devdocs/api/testing_api.html#matplotlib.testing.decorators.image_comparison) looks similar to what you propose except that you have to manually decorate each test and move the reference images manually to a particular location. Your approach sounds more comfortable to use.

However, I am a bit worried about the impact of storing  a lot of images in the git repository. I think it would be better if the images could somehow be stored in a separate location (maybe in the same place as the latest version of the documentation?). Then we would not need to store a lot of slightly different versions of images in the repository, caused by changes in the plotting code (images are much more expensive than text in terms of space).

In any way, please feel free to ping us, as it would be good to know all the alternatives that we have before committing to one.

eliegoudout:
My fixture heavily rlies on `matplotlib.testing` indeed. I personally chose to store plots in `/test/expected/` (so, next to `test/project/`) an indeed it adds up if you have too many plots. I will deinitely update you then!
@ego-thales

examples/representation/plot_mixed_data.py
luisheb: My bad, I did not have pandas-stubs installed

ego-thales:
@vnmabus @luisheb
As promised, I come back to you now that my work project is [opensourced](https://github.com/ThalesGroup/scio) :tada: 

Here's a direct link to `match_plots`
https://github.com/ThalesGroup/scio/blob/837e5c328e2faa4c154bda1146a017cbfe654a11/test/conftest.py#L261C1-L277C8

The use is detailed in the docstring. It has two CLI options (disable or update) and one runtime `ignore()` method to disable locally.

#### What it does
It
- monkeypatches `plt.show` to instead save figures with a normalized name (depending on the test and the number of plots found before in the test), _e.g._ `blablabla.fig0.observed.png`;
- collects expected plots located saved `test/expected/<test_name>/<normalized names>.png`
- compares all `blablabla.fig0.observed.png` with `blablabla.fig0.png`
- Upon mismatch (within very small tolerance I found necessary despite every reproducibility hack), fails the test and also creates the `blablabla.fig0.observed.diff.png` (implemented by `matplotlib` directly). If `--ignore-plots` this is skipped, and if `--update-plots`, then the old `blablabla.fig0.png` is replaced and test is skipped with an update log.

#### Implementation
It is a bit "dirty" in some ways as I did not want to have an infinitely long `conftest.py` 'already long enough..), but should stilll be quite robust.

#### Notes
- If you update multiple plots in the same test, the multiple `pytest.skip` will eventually fail the test (see pytest-dev/pytest/issues/13537), but this does not hurt logic, only report.
- If you `KeyboardInterrupt` during the test, you might end up with an extra `blablabla.fig0.observed.png` file.
- You can also check out the `match_plots_torch` fixture which I use to automatically `ignore()` unless `dtype is torch.float and device == torch.device("cpu")`.
- You can also check out analoguous `match_outerr` for stdout/stderr captures and `match_arrays` which works for numpy ad torch arrays, and can be extended to any class of arrays.
- You can browse my [`test/expected`](https://github.com/ThalesGroup/scio/blob/main/test/expected/test_summary_plot/plot_float-cpu-with_scores_and_layers%3DTrue-with_oods_title%3DTrue-with_hist_kw%3DTrue.fig0.png) folder to get an idea of what it concretely does. But in terms of file size, since `test/` is not shipped with my releases, I do not care too much. Also, it is still quite reasonable.

Finally, I would like to say thanks for you work, since your repository was a good inspiration regarding my own developments. You might find a lot of similarities in the presentation of `scio` in general, it's not by chance!

All the best!

PR #693, opened 2025-06-14
FPCA Basis

luisheb:
FPCA basis added to facilitate its creation without having to use a custom basis.

- [ ] I have added thorough tests for the new/changed functionality


luisheb:
I haven’t added any tests for this new functionality, as it didn’t seem strictly necessary for this implementation. That said, if you think tests should be included, I’m happy to add them.

